PR #1, opened 2023-12-28
Make sure packageinfo object in leapp json configuration has out_packageset/in_packageset before do anything

SandakovMM:


pleskdistup/common/src/leapp_configs.py
```diff
@@ -238,13 +238,19 @@ def add_repositories_mapping(repofiles: typing.List[str], ignore: typing.List =
 
 def set_package_repository(package: str, repository: str, leapp_pkgs_conf_path: str = LEAPP_PKGS_CONF_PATH) -> None:
     pkg_mapping = None
+    log.debug("Reconfigure mapping for package '{}' to repository '{}'".format(package, repository))
```
kpushkaryov, line 241: I suggest to use f-strings or at least named placeholders as above instead of `{}`.

kpushkaryov (changes requested):

kpushkaryov (approved):

PR #6, opened 2024-01-23
Various changes related to ubuntu20to22

kpushkaryov:


pleskdistup/actions/common_checks.py
```diff
@@ -142,3 +144,109 @@ def __init__(self):
...
+
+    @description.setter
+    def description(self, val: str) -> None:
+        self._description = val
+
+    def _do_check(self) -> bool:
+        """Perform the check."""
+        log.debug("Checking minimum free disk space")
+        cmd = [
+            "findmnt", "--output", "source,target,avail",
```
SandakovMM, line 212: Suggest using the complete path to avoid any issues that may arise from users injecting a custom PATH. IIRC I've already meet related problems on centos2alma once.

pleskdistup/actions/common_checks.py
```diff
@@ -142,3 +144,109 @@ def __init__(self):
...
+                stderr=subprocess.PIPE,
+                check=True,
+                universal_newlines=True,
+            )
+            log.debug(
+                f"Command {cmd + [path]} returned {proc.returncode}, "
+                f"stdout: '{proc.stdout}', stderr: '{proc.stderr}'"
+            )
+            fs_data = json.loads(proc.stdout)["filesystems"][0]
+            fs_data["req"] = 0
```
SandakovMM, line 231: Looks like could be moved under `if fs_data["source"] not in filesystems:` condition

pleskdistup/actions/common_checks.py
```diff
@@ -142,3 +144,109 @@ def __init__(self):
...
+        self._description = description
+        self.violations = []
+
+    @property
+    def description(self) -> str:
+        """Description of violations or empty string.
+
+        Can be set to change the message template.
+        """
+        if not self.violations:
```
SandakovMM, line 192: It seems like this situation should never occur. If it does, I would prefer to raise an exception here or, at the very least, return a clear error message such as: "Something has gone very wrong. Please contact us with the feedback archive." Otherwise, there is a chance that the customer may receive information about not meeting conditions without an explanation of the conditions themselves.

pleskdistup/actions/common_checks.py
```diff
@@ -142,3 +144,109 @@ def __init__(self):
...
+            return ""
+        res = self._description
```
SandakovMM, line 194: Suggestion: Do we really need to be able to change `_description` at all? Perhaps we could simply hard-code the initial part here. This way, we can remove the setter and no longer need to check `if res.endswith(":"):`.

pleskdistup/actions/common_checks.py
```diff
@@ -142,3 +144,109 @@ def __init__(self):
...
+                check=True,
+                universal_newlines=True,
+            )
+            log.debug(
+                f"Command {cmd + [path]} returned {proc.returncode}, "
+                f"stdout: '{proc.stdout}', stderr: '{proc.stderr}'"
+            )
+            fs_data = json.loads(proc.stdout)["filesystems"][0]
+            fs_data["req"] = 0
+            fs_data["paths"] = []
```
SandakovMM, line 232: Looks like we could consider using a `set` object here, as we will eventually convert it into a set anyway.

pleskdistup/main.py
```diff
@@ -194,8 +194,9 @@ def do_convert(
...
 ) -> int:
     if not options.resume and not required_conditions_satisfied(upgrader, options, options.phase):
-        printerr("Please fix noted problems before continuing the conversion")
-        return 1
+        printerr("Conversion can't be performed due to the problems noted above")
+        if not show_plan:
+            return 1
 
     actions_map = upgrader.construct_actions(sys.argv[0], options, options.phase)
     dup = find_duplicate_actions(actions_map)
```
SandakovMM, line 202: By the way, not sure we should avoid duplicates. I believe it is fine if we want to perform the same action several times. However it not tied to the changes itself.

pleskdistup/actions/distupgrade.py
```diff
@@ -65,40 +65,49 @@ def estimate_revert_time(self) -> int:
...
+        self.sources_list_path = sources_list_path
+        self.sources_list_d_path = sources_list_d_path
+
+        self._name = name
+
+    @property
+    def name(self):
+        return self._name.format(self=self)
 
     def _change_sources_codename(self, from_codename: str, to_codename: str) -> None:
```
SandakovMM, line 94: Proposal: Maybe we could move this function to some common -> deb specific module and add some unit tests

pleskdistup/actions/distupgrade.py
```diff
@@ -110,6 +119,10 @@ def estimate_revert_time(self) -> int:
         return 20
 
 
+SetupDebianRepositories = SetupAptRepositories
+SetupUbuntuRepositories = SetupAptRepositories
```
SandakovMM, line 123: Nit: It seems that we can remove it and just use `SetupAptRepositories`. However, this means we will need to update the actions for ubuntu18to20 and debian11to12 when we move them to a new version of the submodule.

pleskdistup/main.py
```diff
@@ -272,7 +272,7 @@ def read_resume_data(resume_path: PathType) -> ResumeData:
 
 The utility writes a log to the file specified by --logfile. If there are any issues, you can find more information in the log file.
 
-Utility version is {util_revision}.
+Plesk dist-upgrader version {pleskdistup_revision}.
```
SandakovMM, line 276: Suggest: `Plesk dist-upgrader framework version`

pleskdistup/actions/plesk.py
```diff
@@ -297,6 +297,37 @@ def estimate_revert_time(self) -> int:
...
+        self.name = name
+
+    def _prepare_action(self) -> action.ActionResult:
+        util.logged_check_call([
+            "/usr/sbin/plesk", "bin", "passwords", "--encrypt",
+        ])
+        return action.ActionResult()
+
+    def _post_action(self) -> action.ActionResult:
+        return action.ActionResult(action.ActionState.SKIPPED)
```
SandakovMM, line 316: Please be aware that the use of ActionState.SKIPPED is not meant for an "empty" action. Originally, it was added to skipping of `_post_action` and `_revert_action` when the action isn't needed, based on the result of `_is_required` returning false. While in this case it has no effect, other developers could deside to use SKIPPED in an empty `_prepare_action` could lead to unexpected outcomes.

pleskdistup/common/src/mariadb.py
```diff
@@ -93,12 +93,12 @@ def get_installed_mariadb_version() -> MariaDBVersion:
 
 
 def get_mariadb_config_file_path() -> str:
-    if dist._is_rhel_based(dist.get_distro()):
+    if dist.get_distro().rhel_based:
```
SandakovMM, line 96: I think it would be more natural to write `dist.get_distro().is_rhel_based()` instead. Perhaps we should consider making it a method rather than a parameter.

pleskdistup/actions/common_checks.py
kpushkaryov: This situation would occur normally if the check haven't been performed yet or passed.

pleskdistup/actions/common_checks.py
kpushkaryov: Unfortunately, we can't have just a getter without a setter, because `description` is a base class attribute and you can't override it with non-data descriptor. You will get "Cannot override writeable attribute with read-only property" from mypy. I will just drop the property and assign the attribute directly.

pleskdistup/main.py
kpushkaryov: Here only actions with the same name inside the same stage are considered duplicates. You can have the same name in different stages or choose different names for multiple instances of the same action inside the same stage. This is needed to provide process resuming, where the (stage name, action name) pair is used as an unique key.

pleskdistup/actions/distupgrade.py
kpushkaryov: Actually, the function looks too simple for separate testing. In future it should be replaced with something smarter, because simple string replacement doesn't seem too reliable.

pleskdistup/actions/distupgrade.py
kpushkaryov: This is a precaution if we would need to make them different classes in future. Then `SetupAptRepositories` would be the base class for `SetupDebianRepositories`, `SetupUbuntuRepositories`.

pleskdistup/common/src/mariadb.py
kpushkaryov: Technically it's already a method, just in read-only property form. Since this value can't depend on any parameters, making it an ordinary method and adding `is_` and `()` on any invocation seems redundant to me.

SandakovMM: In my opinion, the issue is more about readability than technicality. I think using 'if distro is rhel based' looks better than 'if distro rhel based' (because of 'is' obviously).
Additionally, I would expect to have a setter for a parameter, or at least it can be naturally added. However, no one will question why I can't add a setter for the `is_rhel_based` method.

pleskdistup/actions/distupgrade.py
SandakovMM: > In future it should be replaced with something smarter, because simple string replacement doesn't seem too reliable.

In this case I'm sure we should add some tests =)

pleskdistup/main.py
SandakovMM: Ok then. However it could complicate maintenance, when actions added by some specific command line options.

pleskdistup/actions/common_checks.py
SandakovMM: Yes, I think it would be better to simply return `_description` in the getter. and set the description from `_do_check`. Additionally, it's not necessary to store valuations in the checker class in this case, which means we store less data and have fewer problems.

pleskdistup/actions/common_checks.py
SandakovMM: I think this shouldn't happen. Descriptions are supposed to explain the problem that was found and provide clear instructions on how to fix it. It seems odd to receive a description of a problem that hasn't been detected yet.

pleskdistup/actions/distupgrade.py
SandakovMM: Anyway, it's just a suggestion. We could just add a github issue and leave it for now.

pleskdistup/common/src/mariadb.py
kpushkaryov: It looks no worse than check of any boolean variable named like `rhel_based`.
Not having a setter is normal for properties, or why else read-only properties were added to Python in the first place? :)

pleskdistup/actions/common_checks.py
kpushkaryov: I'm for results accessibility in machine-readable form, not only human-readable strings.

pleskdistup/actions/common_checks.py
kpushkaryov: Currently unlikely, but it's a pain to work with classes designed for one particular use case in mind, omitting obvious alternatives.
Also, firing an exception in such case won't make user's situation better, because they won't get the description anyways.

SandakovMM: > omitting obvious alternatives.
Could you please provide a obvious example of when it's necessary to set a description outside of a check action?

> pain to work with classes designed for one particular use case
There is always a tradeoff between 'pain to work with' and 'pain to maintain'. I believe a class should do exactly the things it was destined for. Otherwise we could add encryption into checker because "it could be useful, why not".

> Also, firing an exception in such case won't make user's situation better, because they won't get the description anyways.
IIRC we will show an error message in this case. Anyway we should have clear description what should be done by customer to solve the problem.

Additionally, I'm not sure other check classes are ready for such way of use

pleskdistup/actions/common_checks.py
SandakovMM: > I'm for results accessibility in machine-readable form, not only human-readable strings.

Hm, it might be better to make it possible to return pre-checker results in json (or any similar format)

pleskdistup/actions/common_checks.py
kpushkaryov: I'm not talking about setting the description outside of a check action, I'm just saying that the description can be legitimately empty, if the check hasn't been performed yet or passed. Please not that the property code has been dropped in further commits in favor of direct assignment to the `description` attribute.

pleskdistup/actions/common_checks.py
kpushkaryov: At least it would be nice if checks would provide access to unserialized results for potential use in other parts of the program, e.g. for testing.

pleskdistup/common/src/mariadb.py
SandakovMM: > It looks no worse than check of any boolean variable named like rhel_based.

Yep, it looks worse then checking `is_rhel_based` method =)

> Not having a setter is normal for properties, or why else read-only properties were added to Python in the first place? :)

As far as I know we can't prevent from adding setter for property and can't evidently show it should be read-only. I might be wrong at this.

kpushkaryov: If you want to prevent someone from doing something, Python is not the language you're looking for :)
For example, anyone can add `set_is_rhel_based(x)` later or do `obj.is_rhel_based = foobar` either.
But assignment to a read-only property would result in an error, so you can't inadvertently overwrite it.

kpushkaryov: Some insane person can even do `obj.is_rhel_based = lambda: return False` :)

SandakovMM: > Some insane person can even do obj.is_rhel_based = lambda: return False :)

I will remember it =)
Anyway it's all about a matter of taste, so let's skip it.

pleskdistup/actions/common_checks.py
SandakovMM: > e.g. for testing.

I believe in this case you should split function that takes results and test this new function.

SandakovMM (approved):

PR #8, opened 2024-02-13
Introduce centos2alma support

SandakovMM:


pleskdistup/actions/extensions.py
```diff
@@ -32,3 +37,88 @@ def estimate_post_time(self) -> int:
...
+        self.description = "rebundling ruby applications"
+
+    def _is_ruby_domain(self, domain_path) -> bool:
+        return os.path.exists(os.path.join(domain_path, ".rbenv"))
+
+    def _is_required(self) -> bool:
+        if not os.path.exists("/var/lib/rbenv/versions/"):
+            return False
+
+        return any(self._is_ruby_domain(domain) for domain in os.scandir("/var/www/vhosts"))
```
kpushkaryov, line 59: `/var/www/vhosts` also contains non-domain subdirectories like `chroot` or `system`.

pleskdistup/actions/extensions.py
```diff
@@ -32,3 +37,88 @@ def estimate_post_time(self) -> int:
 
     def estimate_revert_time(self) -> int:
         return 20
+
+
+# We should do rebundling of ruby applications after the conversion
+# because some of our libraries will be missing.
+# The prime example of missing library - libmysqlclient.so.18 required by mysql2 gem
+class RebundleRubyApplications(action.ActiveAction):
+    PLESK_APACHE_CONFIGS_PATH = "/etc/httpd/conf/plesk.conf.d/vhosts/"
```
kpushkaryov, line 46: Doesn't work for Debian-based systems.

pleskdistup/actions/extensions.py
```diff
@@ -32,3 +37,88 @@ def estimate_post_time(self) -> int:
...
+
+    def _prepare_action(self) -> action.ActionResult:
+        return action.ActionResult()
+
+    def _get_ruby_application_path_by_config(self, domain_name: str) -> str:
+        apache_config = os.path.join(self.PLESK_APACHE_CONFIGS_PATH, domain_name + ".conf")
+
+        if not os.path.isfile(apache_config) or not files.find_file_substrings(apache_config, "PassengerRuby"):
+            # Likely it means ruby application is not enabled, so we need to do direct search in the filesystem
+            return None
```
kpushkaryov, line 69: Please do type checks. This function is annotated as strictly returning `str`.

pleskdistup/actions/extensions.py
```diff
@@ -32,3 +37,88 @@ def estimate_post_time(self) -> int:
...
+        application_root = files.find_file_substrings(apache_config, "PassengerAppRoot")
+        if not application_root:
+            return None
+
+        # Record format is: PassengerAppRoot "[path]"
+        # We can expect it is always in this format since configuration is generated by plesk
+        return application_root[0].split()[1].strip("\"")
+
+    def _get_ruby_application_path_by_bundle_subdir(self, domain_home: str) -> str:
+        bundle = files.find_subdirectory_by(domain_home,
```
kpushkaryov, line 80: Please put a newline after the left parenthesis.

pleskdistup/actions/extensions.py
```diff
@@ -32,3 +37,88 @@ def estimate_post_time(self) -> int:
...
+                                            lambda subdir: os.path.basename(subdir) == "bundle" and
+                                                           os.path.exists(os.path.join(subdir, "ruby")))
+        if bundle is None or not os.path.isdir(bundle):
+            return None
+
+        return os.path.dirname(os.path.dirname(bundle))
+
+    def _get_ruby_application_paths(self, domain_path: str) -> typing.Tuple[str, str]:
+        app_directory = self._get_ruby_application_path_by_config(domain_path.name)
```
kpushkaryov, line 89: `domain_path` is declared as `str`. Please do type checks here and elsewhere.

pleskdistup/actions/extensions.py
```diff
@@ -32,3 +37,88 @@ def estimate_post_time(self) -> int:
...
+            log.debug("Ruby application is disabled. Search for application root in filesystem. Domain home directory: {}".format(domain_path))
+            app_directory = self._get_ruby_application_path_by_bundle_subdir(domain_path)
+
+        if app_directory is None:
+            return None, None
+
+        return app_directory, os.path.join(app_directory, "vendor", "bundle")
+
+    def _post_action(self) -> action.ActionResult:
+        ruby_domains = (domain_path for domain_path in os.scandir("/var/www/vhosts") if self._is_ruby_domain(domain_path))
```
kpushkaryov, line 100: See above about `/var/www/vhosts`.

pleskdistup/actions/extensions.py
```diff
@@ -32,3 +37,88 @@ def estimate_post_time(self) -> int:
...
+                                                           os.path.exists(os.path.join(subdir, "ruby")))
+        if bundle is None or not os.path.isdir(bundle):
+            return None
+
+        return os.path.dirname(os.path.dirname(bundle))
+
+    def _get_ruby_application_paths(self, domain_path: str) -> typing.Tuple[str, str]:
+        app_directory = self._get_ruby_application_path_by_config(domain_path.name)
+        if app_directory is None:
+            log.debug("Ruby application is disabled. Search for application root in filesystem. Domain home directory: {}".format(domain_path))
```
kpushkaryov, line 91: Please use f-strings here and below.

pleskdistup/actions/extensions.py
```diff
@@ -32,3 +37,88 @@ def estimate_post_time(self) -> int:
...
+            shutil.rmtree(bundle)
+            util.logged_check_call(["/usr/sbin/plesk", "sbin", "rubymng", "run-bundler", username, app_directory])
+
+        return action.ActionResult()
+
+    def _revert_action(self) -> action.ActionResult:
+        return action.ActionResult()
+
+    def estimate_post_time(self) -> int:
+        return 60 * len([domain_path for domain_path in os.scandir("/var/www/vhosts") if self._is_ruby_domain(domain_path)])
```
kpushkaryov, line 124: Listing Ruby domains probably should be extracted into a function, because it's repeated multiple times above.

pleskdistup/actions/spamassassin.py
```diff
@@ -31,3 +33,67 @@ def _post_action(self) -> action.ActionResult:
...
+        return action.ActionResult()
+
+    def _post_action(self) -> action.ActionResult:
+        util.logged_check_call(["/usr/sbin/plesk", "sbin", "spammng", "--enable"])
+        util.logged_check_call(["/usr/sbin/plesk", "sbin", "spammng", "--update", "--enable-server-configs", "--enable-user-configs"])
+
+        util.logged_check_call(["/usr/bin/systemctl", "daemon-reload"])
+        util.logged_check_call(["/usr/bin/systemctl", "enable", "spamassassin.service"])
+
+        if rpm.handle_rpmnew(SPAMASSASIN_CONFIG_PATH):
```
kpushkaryov, line 59: What about Debian-based systems?

pleskdistup/actions/spamassassin.py
```diff
@@ -31,3 +33,67 @@ def _post_action(self) -> action.ActionResult:
...
+            motd.add_finish_ssh_login_message("Note that spamassasin configuration '{}' was changed during conversion. "
```
kpushkaryov, line 60: Please put a newline after the left parenthesis and use f-strings.

pleskdistup/actions/spamassassin.py
```diff
@@ -31,3 +33,67 @@ def _post_action(self) -> action.ActionResult:
     def _revert_action(self) -> action.ActionResult:
         os.unlink(self.spamassasin_backup_path)
         return action.ActionResult()
+
+
+class HandleUpdatedSpamassassinConfig(action.ActiveAction):
+    # Make sure the trick is preformed before any call of 'systemctl daemon-reload'
+    # because we change spamassassin.service configuration in scope of this action.
+    def __init__(self):
+        self.name = "fix spamassassin configuration"
```
kpushkaryov, line 42: This string doesn't correspond to class name.

pleskdistup/actions/spamassassin.py
```diff
@@ -31,3 +33,67 @@ def _post_action(self) -> action.ActionResult:
...
+
+    def _revert_action(self) -> action.ActionResult:
+        util.logged_check_call(["/usr/bin/systemctl", "enable", "spamassassin.service"])
+        util.logged_check_call(["/usr/bin/systemctl", "start", "spamassassin.service"])
+        return action.ActionResult()
+
+
+class AssertSpamassassinAdditionalPluginsDisabled(action.CheckAction):
+    def __init__(self):
+        self.name = "check spamassassin additional plugins enabled"
```
kpushkaryov, line 72: So, enabled or disabled?

pleskdistup/actions/systemd.py
```diff
@@ -75,3 +75,100 @@ def _post_action(self) -> action.ActionResult:
     def _revert_action(self) -> action.ActionResult:
         systemd.remove_systemd_service(self.service_name)
         return action.ActionResult()
+
+
+class RulePleskRelatedServices(action.ActiveAction):
```
kpushkaryov, line 80: It's not clear what "rule" means here.

pleskdistup/actions/systemd.py
```diff
@@ -75,3 +75,100 @@ def _post_action(self) -> action.ActionResult:
...
+            "plesk-task-manager.service",
+            "plesk-web-socket.service",
+            "sw-cp-server.service",
+            "sw-engine.service",
+        ]
+        self.plesk_basic_services = [service for service in self.plesk_basic_services if systemd.is_service_exists(service)]
+
+    def _enable_services(self) -> action.ActionResult:
+        # MariaDB could be started before, so we should stop it first
+        # ToDo. Or we could check it is started and just remove it from list
```
kpushkaryov, line 158: TODO

pleskdistup/common/src/action.py
```diff
@@ -390,8 +390,8 @@ def __exit__(self, *kwargs):
 
     def _get_flow(self) -> typing.Dict[str, typing.List[ActiveAction]]:
         res = {}
-        for stage_id, actions in reversed(self.stages.items()):
-            res[stage_id] = list(reversed(actions))
+        for stage_id, actions in reversed(list(self.stages.items())):
```
kpushkaryov, line 393: It's a temporary iterator, why use `list` here?

pleskdistup/common/src/action.py
```diff
@@ -390,8 +390,8 @@ def __exit__(self, *kwargs):
...
+            res[stage_id] = list(reversed(list(actions)))
```
kpushkaryov, line 394: Why additional `list`?

pleskdistup/messages.py
```diff
@@ -55,3 +55,10 @@
 """
 
 NOT_SUPPORTED_ERROR = "Your distribution is not supported yet, please contact Plesk support for further assistance."
+
+REBOOT_WARN_MESSAGE = """\r\033[93m****************************** WARNING ***********************************************
+\033[92mThe conversion is ready to begin. The server will be rebooted in {delay} seconds.
+The conversion process will take approximately 25 minutes. If you wish to prevent the reboot, simply
+terminate the centos2alma process. Please note that Plesk functionality is currently unavailable.
```
kpushkaryov, line 62: Please don't hard-code "centos2alma" here.

kpushkaryov (changes requested):
Please ensure type correctness. I advise to run `mypy` for type checks and `flake8` for general checks.

SandakovMM:
> Please ensure type correctness. I advise to run `mypy` for type checks and `flake8` for general checks.

It looks like we should a github action for this

pleskdistup/actions/common.py
```diff
@@ -224,3 +226,77 @@ def estimate_post_time(self):
 
     def estimate_revert_time(self):
         return 0
+
+
+class RuleSelinux(action.ActiveAction):
```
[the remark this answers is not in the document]
SandakovMM: Hm. it not really makes it permissive, since we return enforce mode after the conversion. I propose `DisableSelinuxDuringPerformance` or something like this.

pleskdistup/actions/common.py
```diff
@@ -224,3 +226,77 @@ def estimate_post_time(self):
...
+        return action.ActionResult()
+
+    def _post_action(self) -> action.ActionResult:
+        return action.ActionResult()
+
+    def _revert_action(self) -> action.ActionResult:
+        return action.ActionResult()
+
+
+class FixSyslogLogrotateConfig(action.ActiveAction):
```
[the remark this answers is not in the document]
SandakovMM: Might be, let's move this out

pleskdistup/actions/common_checks.py
```diff
@@ -94,6 +98,170 @@ def _do_check(self) -> bool:
...
+    # TODO: get rid of the explicit version list
+    return [version.PHPVersion(ver) for ver in (
+                "5.2", "5.3", "5.4", "5.5", "5.6",
+                "7.0", "7.1", "7.2", "7.3", "7.4",
+                "8.0", "8.1", "8.2", "8.3",
+                )
+            ]
+
+
+class AssertOutdatedPhpVersionNotInstalled(action.CheckAction):
```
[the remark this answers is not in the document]
SandakovMM: I think the issue with this name is that it doesn't explain why we're checking these versions. It seems like we just arbitrarily chose the minimum version. In other words, I'd prefer to include the word "outdated" in the name of the action to better reflect its purpose.

pleskdistup/actions/common_checks.py
```diff
@@ -94,6 +98,170 @@ def _do_check(self) -> bool:
...
+
+        self.description = self.description.format(
+            ", ".join([outdated_php_packages[installed] for installed in installed_pkgs]),
+            " ".join(outdated_php_packages[installed].replace(" ", "") for installed in installed_pkgs).lower())
+
+        log.debug("Outdated PHP versions found")
+        return False
+
+
+class AssertWebsitesUsesOutdatedPHP(action.CheckAction):
```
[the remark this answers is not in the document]
SandakovMM: Mostly right, however I still prefer to include outdated in the name

pleskdistup/actions/common_checks.py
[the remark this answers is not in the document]
kpushkaryov: I think, the name of an assertion should clearly state what it checks, not why it checks. "Why" is a matter of context. Like `fopen()` means you open a file, not why you open it.

It's OK to create a purpose-specific alias in a particular upgrader, e.g. `AssertOutdatedPhpVersionNotInstalled = AssertMinPhpVersionInstalled`, but dist-upgrader itself should be more generic.

It's also not clear what an assertion `AssertOutdatedPhpVersionNotInstalled(X)` checks, when you see it in code. Judging by its name, it fails if PHP X is installed (actually, it's not true currently). But will it fail if PHP X-1 is installed? Not clear from the name and since there's no docstring, you need to read the code to find that.

pleskdistup/actions/common_checks.py
```diff
@@ -94,6 +98,170 @@ def _do_check(self) -> bool:
...
+            if len(outdated_php_domains) <= 0:
+                return True
+
+            log.debug(f"Outdated PHP domains: {outdated_php_domains}")
+            outdated_php_domains = "\n\t- ".join(outdated_php_domains)
+            self.description = self.description.format(self.last_outdated_version, outdated_php_domains)
+
+            return False
+        except Exception:
+            log.error("Unable to get domains list from plesk database!")
```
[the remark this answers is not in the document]
SandakovMM: Yes, because this issue isn't directly related to our checker, and we can't confirm if any website is using outdated PHP in this case. Actually, we could simply raise an exception again here. However, there might be a problem if a previous tool call has disabled the database, I should check it.

pleskdistup/actions/common_checks.py
```diff
@@ -94,6 +98,170 @@ def _do_check(self) -> bool:
...
+
+        return True
+
+
+class AssertCronUsesOutdatedPHP(action.CheckAction):
```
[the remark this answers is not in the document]
SandakovMM: Same commentary about outdated, as in other similar places

pleskdistup/actions/common_checks.py
```diff
@@ -94,6 +98,170 @@ def _do_check(self) -> bool:
...
+        self.description = """We have detected that some cronjobs are using outdated PHP versions.
+\tSwitch the following cronjobs to to any version higher than {} in order to continue with the conversion process:"
+\t- {}
+
+\tYou can do this in the Plesk web interface by going “Tools & Settings” → “Scheduled Tasks”.
+"""
+
+    def _do_check(self) -> bool:
+        outdated_php_packages = [
+            f"plesk-php{php.major}{php.minor}" for php in get_known_php_versions() if php < self.last_outdated_version
```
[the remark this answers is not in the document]
SandakovMM: Maybe we even could move it to common part, and add some little tests

pleskdistup/actions/common_checks.py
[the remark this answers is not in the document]
SandakovMM: However, since the behavior is a little different and I want to ensure the update of distupgraders goes smoothly, it seems necessary to verify the exact minimal version. Let's hope description of the action is well enough in this case.

pleskdistup/actions/extensions.py
SandakovMM: Yep, but we don't expect to find .rbenv directory in it. However we could simply exclude this two directories

pleskdistup/actions/spamassassin.py
SandakovMM: This action will be just skipped on deb-based. So I believe we could just add TO-DO here if someone would like to use the action on deb-based.

pleskdistup/actions/systemd.py
SandakovMM: The same thing as for se-linux actually. Let rename this action a little, maybe this will help

pleskdistup/common/src/action.py
SandakovMM: Because it seems like python 3.6 doesn't support `reversed` for dict_items. Described in the commit message

pleskdistup/common/src/action.py
SandakovMM: Same answer as for previous commentary

pleskdistup/actions/extensions.py
kpushkaryov: What about this?

pleskdistup/actions/extensions.py
```diff
@@ -97,20 +111,19 @@ def _get_ruby_application_paths(self, domain_path: str) -> typing.Tuple[str, str
...
-        for domain_path in ruby_domains:
-            log.debug("Re-bundling ruby application in domain: {}".format(domain_path.name))
+        for domain_path in self._get_ruby_domains():
+            log.debug(f"Re-bundling ruby application in domain: {domain_path.name}")
 
             app_directory, bundle = self._get_ruby_application_paths(domain_path)
 
             if bundle is None or not os.path.isdir(bundle):
-                log.debug("Skip re-bundling for non bundling domain '{}'".format(domain_path.name))
+                log.debug(f"Skip re-bundling for non bundling domain '{domain_path.name}'")
```
kpushkaryov, line 120: No `domain_path.name`.

pleskdistup/actions/spamassassin.py
```diff
@@ -36,10 +36,12 @@ def _revert_action(self) -> action.ActionResult:
 
 
 class HandleUpdatedSpamassassinConfig(action.ActiveAction):
+    name: str
```
kpushkaryov, line 39: Redeclaration.

pleskdistup/actions/spamassassin.py
kpushkaryov: Do we have a check anywhere to skip it?

pleskdistup/actions/systemd.py
```diff
@@ -77,7 +77,10 @@ def _revert_action(self) -> action.ActionResult:
         return action.ActionResult()
 
 
-class RulePleskRelatedServices(action.ActiveAction):
+class DisablePleskRelatedServicesDuringPerformance(action.ActiveAction):
+    name: str
```
kpushkaryov, line 81: Redeclaration.

pleskdistup/actions/systemd.py
```diff
@@ -77,7 +77,10 @@ def _revert_action(self) -> action.ActionResult:
         return action.ActionResult()
 
 
-class RulePleskRelatedServices(action.ActiveAction):
+class DisablePleskRelatedServicesDuringPerformance(action.ActiveAction):
```
kpushkaryov, line 80: Performance -> Upgrade

pleskdistup/messages.py
```diff
@@ -59,6 +59,6 @@
 REBOOT_WARN_MESSAGE = """\r\033[93m****************************** WARNING ***********************************************
```
kpushkaryov, line 59: Where is it used?

pleskdistup/actions/common.py
```diff
@@ -228,7 +228,11 @@ def estimate_revert_time(self):
         return 0
 
 
-class RuleSelinux(action.ActiveAction):
+class DisableSelinuxDuringPerformance(action.ActiveAction):
```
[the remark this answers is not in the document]
SandakovMM: It's not always upgrade, sometimes conversion. But I suppose it looks fine anyway

pleskdistup/actions/common_checks.py
```diff
@@ -98,40 +98,28 @@ def _do_check(self) -> bool:
...
     def __init__(
         self,
-        last_outdated_version: str,
+        min_version: str,
     ):
         self.name = "check for outdated PHP versions"
-        self.last_outdated_version = version.PHPVersion(last_outdated_version)
-        self.description = """Outdated PHP versions were detected: '{}'.
+        self.min_version = version.PHPVersion(min_version)
+        self.description = """Outdated PHP versions were detected: '{versions}'.
```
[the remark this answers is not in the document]
SandakovMM: Actually ':' is enough, so let's remove quotes

pleskdistup/actions/common_checks.py
```diff
@@ -142,124 +130,106 @@ def _do_check(self) -> bool:
...
 
     def _do_check(self) -> bool:
-        outdated_php_packages = [
-            f"plesk-php{php.major}{php.minor}" for php in get_known_php_versions() if php < self.last_outdated_version
-        ]
-        log.debug(f"Outdated PHP handlers: {outdated_php_packages}")
+        log.debug(f"Checking for outdated PHP version in cronjobs.  {self.min_version}")
+        if not plesk.is_plesk_database_ready():
+            log.info("Plesk database is not ready. Skipping the outdated php for websites check.")
+            return True
```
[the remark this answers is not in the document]
SandakovMM: The conversion process does not rely heavily on the Plesk database. I don't think the distupgrade process does either.
Additionally, I believe the PHP checker is not the appropriate place to check for the availability of the Plesk database. We could include a separate CheckAction specifically for this purpose.

pleskdistup/messages.py
SandakovMM: I am passing this message to the `PreRebootPause` action. However it could take any other message as well.

pleskdistup/actions/spamassassin.py
SandakovMM: We are looking for a .rpmnew file inside the function. However, we could add a check to make sure it is rpm-based system.

pleskdistup/actions/common_checks.py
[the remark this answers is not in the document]
kpushkaryov: Ignoring the check if it can't be performed due to technical problems seems unsafe as default behaviour. By default, the answer must be yes, no or hard fail. I propose to make the soft fail opt-in in such cases, e.g. `AssertMinPhpVersionUsedByCron(optional=True)`.

kpushkaryov (changes requested):

pleskdistup/actions/spamassassin.py
```diff
@@ -31,3 +33,74 @@ def _post_action(self) -> action.ActionResult:
...
+        util.logged_check_call(["/usr/sbin/plesk", "sbin", "spammng", "--enable"])
+        util.logged_check_call(["/usr/sbin/plesk", "sbin", "spammng", "--update", "--enable-server-configs", "--enable-user-configs"])
+
+        util.logged_check_call(["/usr/bin/systemctl", "daemon-reload"])
+        util.logged_check_call(["/usr/bin/systemctl", "enable", "spamassassin.service"])
+
+        # TODO. Following action is not supported on deb-based system. Actually it will be just skipped.
+        # So if you are going to use the action on deb-based, you should be ready there will be no .rpmnew
+        # things or even file here (obviously).
+        if dist.get_distro().rhel_based():
```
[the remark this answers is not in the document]
SandakovMM: It depends on changes in the spamassasin package. If I recall correctly, we should make these changes because we need to add plesk configurations to the new configuration file.
This is the main purpose of the action. Therefore, if we do not need to do this, we should not use the action at all.

pleskdistup/actions/common_checks.py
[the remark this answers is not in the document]
SandakovMM: It depends on whether we see the unreachable database as a "technical problem" or not. In the case of conversion/distupgrade, it can be a normal state of the system because the tool disables the Plesk database by itself.
However, I'm okay with it as long as it is optional by default.

pleskdistup/common/src/plesk.py
```diff
@@ -166,7 +166,7 @@ def is_plesk_database_ready() -> bool:
 
 
 def get_from_plesk_database(query: str) -> typing.List[str]:
-    if is_plesk_database_ready():
+    if not is_plesk_database_ready():
         # This could be fine when we just restart the conversion/distupgrade tool
         # However, let's log this anyway, it might be a good point to reveal problems
         log.warn("Plesk database is not ready")
```
kpushkaryov, line 172: I suggest to at least return `None` instead of `[]` to distinguish between failure and empty result.

kpushkaryov (dismissed):
Please rebase before merge.

kpushkaryov (approved):

PR #10, opened 2024-02-15
Restore original message of the day, before making in progress message

SandakovMM:
Since process has been started from scratch, we should remove previous in-progress message and fill it from the scratch


kpushkaryov:
I'm curious, in which case did you encounter restart from scratch? Done actions should be skipped by `--resume`.

SandakovMM:
Customers may need to call the script again if it fails. For instance, customers using centos2alma successfully fixed issues with peapp by themselves.Then they just ran the script again to complete the conversion.

kpushkaryov (approved):

PR #14, opened 2024-03-04
Introduce action to increase dovecot DH parameters size

SandakovMM:
Originally required by centos2alma since we upgrade rouncube during the converion process

### DIFF
--- a/pleskdistup/actions/emails.py
+++ b/pleskdistup/actions/emails.py
@@ -1,0 +1,54 @@
+# Copyright 2023-2024. WebPros International GmbH. All rights reserved.
+import json
+import subprocess
+
+from pleskdistup.common import action, log, util
+
+
+class SetMinDovecotDhParamSize(action.ActiveAction):
+    dhparam_size: int
+
+    def __init__(self, dhparam_size: int):
+        self.name = "increase Dovecot DH parameters size to 2048 bits"
+        self.dhparam_size = dhparam_size
+
+    def _is_required(self) -> bool:
+        proc = subprocess.run(
+            ["/usr/sbin/plesk", "sbin", "sslmng", "--show-config"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+        )
+        if proc.returncode != 0:
+            log.warn(f"Failed to get ssl configuration by plesk sslmng: {proc.stdout}\n{proc.stderr}")
+            return False
+
+        try:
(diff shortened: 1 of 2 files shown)


pleskdistup/actions/emails.py
```diff
@@ -0,0 +1,52 @@
+# Copyright 2023-2024. WebPros International GmbH. All rights reserved.
+import json
+import subprocess
+
+from pleskdistup.common import action, log, util
+
+
+class IncreaseDovecotDHParameters(action.ActiveAction):
```
kpushkaryov, line 8: For the sake of generality I propose to call that like `SetMinDovecotDhParamSize` and make `dhparam_size` configurable.

pleskdistup/actions/emails.py
```diff
@@ -0,0 +1,52 @@
...
+    def _is_required(self) -> bool:
+        proc = subprocess.run(
+            ["/usr/sbin/plesk", "sbin", "sslmng", "--show-config"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True,
+        )
+        if proc.returncode != 0:
+            log.warn(f"Failed to get ssl configuration by plesk sslmng: {proc.stdout}\n{proc.stderr}")
+            return False
```
kpushkaryov, line 22: Do we really want to skip it in case of failure?

pleskdistup/actions/emails.py
```diff
@@ -0,0 +1,52 @@
...
+import json
+import subprocess
+
+from pleskdistup.common import action, log, util
+
+
+class IncreaseDovecotDHParameters(action.ActiveAction):
+    def __init__(self):
+        self.name = "increase Dovecot DH parameters size to 2048 bits"
+        self.dhparam_size = 2048
```
kpushkaryov, line 11: Please declare `dhparam_size` type.

kpushkaryov (changes requested):

pleskdistup/actions/emails.py
kpushkaryov: What about that?

kpushkaryov (changes requested):

pleskdistup/actions/emails.py
SandakovMM: Previous commentary sucked as pending. Please take a look

kpushkaryov (approved):

kpushkaryov:
Please rebase before merge.

PR #15, opened 2024-03-05
Handle kernel versions with '_' before build number

SandakovMM:


pleskdistup/common/tests/versiontests.py
```diff
@@ -34,6 +34,9 @@ def test_kernel_start_with_prefix(self):
     def test_kernel_start_with_plus_prefix(self):
         self._check_parse("kernel-plus-3.10.0-327.36.3.el7.centos.plus.x86_64", "3.10.0-327.36.3.el7.centos.plus.x86_64")
 
+    def test_kernel_with_underline(self):
+        self._check_parse("kernel-3.14.43_1-2.x86_64", "3.14.43-1.2.x86_64")
```
kpushkaryov, line 38: The versions are getting convoluted. I propose to test components of `KernelVersion` separately. That is, in addition to testing string representation, also test that `major`, `minor`, etc. are correct.

kpushkaryov (changes requested):

pleskdistup/common/tests/versiontests.py
SandakovMM: > The versions are getting convoluted.
Yep. Unfortunately, I don't think we could do much about it.
Let's add some test here as you proposed.

pleskdistup/common/tests/versiontests.py
```diff
@@ -34,6 +34,16 @@ def test_kernel_start_with_prefix(self):
     def test_kernel_start_with_plus_prefix(self):
         self._check_parse("kernel-plus-3.10.0-327.36.3.el7.centos.plus.x86_64", "3.10.0-327.36.3.el7.centos.plus.x86_64")
 
+    def test_kernel_with_underline(self):
+        kernel = version.KernelVersion("kernel-3.14.43_1-2.x86_64")
+        self.assertEqual(str(kernel), "3.14.43-1.2.x86_64")
+        self.assertEqual(kernel.major, "3")
```
kpushkaryov, line 40: I suggest to expand other tests in the same way later.

kpushkaryov (approved):

pleskdistup/common/tests/versiontests.py
kpushkaryov: Please rebase before merge.

SandakovMM:
@kpushkaryov Could you please approve once again. Changes from other commit leaked into the branch, so I have to remove it and re-base once again.

kpushkaryov (approved):

PR #16, opened 2024-03-12
Add an action to verify if certain websites use PHP from the operating system vendor

SandakovMM:
It is important to avoid conversion in this case because the version of PHP provided by the vendor is 5.4 on CentOS 7 and 7.2 on AlmaLinux 8. After the conversion, there is uncertainty whether the websites will function properly due to the non-trivial task of migrating from PHP 5.4 to PHP 7.2

### DIFF
--- a/pleskdistup/actions/common_checks.py
+++ b/pleskdistup/actions/common_checks.py
@@ -242,6 +242,55 @@ class AssertMinPhpVersionUsedByCron(action.CheckAction):
         return False
 
 
+class AssertOsVendorPhpUsedByWebsites(action.CheckAction):
+    min_version: version.PHPVersion
+
+    def __init__(
+            self,
+            min_version: str,
+    ):
+        self.name = "checking OS vendor PHP used by websites"
+        self.min_version = version.PHPVersion(min_version)
+        self.description = """We have detected that some domains are using the OS vendor PHP version.
+\tSwitch the following domains to {modern} or later in order to continue with the conversion process:
+\t- {domains}
+
+\tYou can achieve this by executing the following command:
+\t> plesk bin domain -u [domain] -php_handler_id plesk-php80-fastcgi
+"""
+
+    def _do_check(self) -> bool:
+        log.debug("Checking the OS vendor PHP version used by the websites")
+        if not plesk.is_plesk_database_ready():
+            log.info("Plesk database is not ready. Skipping the OS vendor PHP check.")
+            return True
+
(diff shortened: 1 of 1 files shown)


pleskdistup/actions/common_checks.py
```diff
@@ -242,6 +242,54 @@ def _do_check(self) -> bool:
         return False
 
 
+class AssertOSVendorPHPUsedByWebsites(action.CheckAction):
```
kpushkaryov, line 245: Please capitalize only the first letter of each word, like `AssertMinPhpVersionUsedByCron` above.

pleskdistup/actions/common_checks.py
```diff
@@ -242,6 +242,54 @@ def _do_check(self) -> bool:
...
+\t> plesk bin domain -u [domain] -php_handler_id plesk-php80-fastcgi
+"""
+
+    def _do_check(self) -> bool:
+        log.debug("Checking the OS vendor PHP version used by the websites")
+        if not plesk.is_plesk_database_ready():
+            log.info("Plesk database is not ready. Skipping the OS vendor PHP check.")
+            return True
+
+        os_vendor_php_handlers = ["'fpm'", "'fastcgi'"]
```
kpushkaryov, line 268: Propose to add inner single quotes dynamically later instead of typing them here.

pleskdistup/actions/common_checks.py
```diff
@@ -242,6 +242,54 @@ def _do_check(self) -> bool:
...
+                return True
+
+            log.debug(f"OS vendor PHP domains: {os_vendor_php_domains}")
+            os_vendor_php_domains = "\n\t- ".join(os_vendor_php_domains)
+            self.description = self.description.format(
+                modern=self.min_version,
+                domains=os_vendor_php_domains
+            )
+        except Exception as ex:
+            log.err("Unable to get domains list from plesk database!")
```
kpushkaryov, line 287: I think, this and the following message are not informative enough. We were not asking about a generic list of domains, but about the list with particular PHP handlers.

kpushkaryov (changes requested):

pleskdistup/actions/common_checks.py
SandakovMM: Hm, seems like we could simply add them dynamically in place

kpushkaryov (approved):
Please rebase before merge.

PR #17, opened 2024-03-13
Add an action to restore original grub configuration on revert

SandakovMM:


pleskdistup/actions/common.py
```diff
@@ -277,3 +277,44 @@ def _post_action(self) -> action.ActionResult:
 
     def _revert_action(self) -> action.ActionResult:
         return action.ActionResult()
+
+
+class RevertChangesInGrub(action.ActiveAction):
+    grub_configs_paths: typing.List[str]
+
+    def __init__(self):
+        self.name = "revert changes in grub made by elivate"
```
kpushkaryov, line 286: grub -> GRUB
elivate -> ELevate

pleskdistup/actions/common.py
```diff
@@ -277,3 +277,44 @@ def _post_action(self) -> action.ActionResult:
...
+        self.grub_configs_paths = [
+            "/boot/grub2/grub.cfg",
```
kpushkaryov, line 288: Are these all the possible paths? E.g. I see `/boot/grub` instead of `/boot/grub2` on Ubuntu 22.04.

kpushkaryov (changes requested):

pleskdistup/actions/common.py
SandakovMM: Sure, feel free to propose other paths as well

kpushkaryov (approved):
Please rebase before merge.

PR #18, opened 2024-03-15
Add an action to restore /ect/dovecot/dovecto.conf to state before conversion/distupgrade

SandakovMM:


pleskdistup/actions/emails.py
```diff
@@ -52,3 +54,31 @@ def _revert_action(self) -> action.ActionResult:
...
+
+    def _prepare_action(self) -> action.ActionResult:
+        files.backup_file(self.dovecot_config_path)
+        return action.ActionResult()
+
+    def _post_action(self) -> action.ActionResult:
+        path_to_backup = plesk.CONVERTER_TEMP_DIRECTORY + "/dovecot.conf.bak"
+        if os.path.exists(self.dovecot_config_path):
+            shutil.copy(self.dovecot_config_path, path_to_backup)
+            motd.add_finish_ssh_login_message(f"The dovecot configuration '{self.dovecot_config_path}' has been restored from CentOS 7. Modern configuration was placed in '{path_to_backup}'.")
```
kpushkaryov, line 77: Does anything confines this action to CentOS 7 only? If not, propose to drop CentOS 7 mention from the message.

pleskdistup/actions/emails.py
```diff
@@ -52,3 +54,31 @@ def _revert_action(self) -> action.ActionResult:
...
+
+    def _is_required(self) -> bool:
+        return os.path.exists(self.dovecot_config_path)
+
+    def _prepare_action(self) -> action.ActionResult:
+        files.backup_file(self.dovecot_config_path)
+        return action.ActionResult()
+
+    def _post_action(self) -> action.ActionResult:
+        path_to_backup = plesk.CONVERTER_TEMP_DIRECTORY + "/dovecot.conf.bak"
```
kpushkaryov, line 74: What is `plesk.CONVERTER_TEMP_DIRECTORY`? Can't find it's definition.
Also, please use `os.path.join()`.

kpushkaryov (changes requested):
Also, typos in the commit message: "ect", "dovecto.conf".

pleskdistup/actions/emails.py
SandakovMM: I think could be used for other repositories as well, so lets remove mention of centos 7

pleskdistup/actions/emails.py
SandakovMM: > What is plesk.CONVERTER_TEMP_DIRECTORY?

It should be passed as argument, actually it is state_dir.

kpushkaryov:
Please fix commit message typos before merge.

kpushkaryov (approved):

PR #19, opened 2024-03-20
Add function to check if service can be started

SandakovMM:
Based on 2 facts:
- service exists
- all required services are exist as well

### DIFF
--- a/pleskdistup/actions/systemd.py
+++ b/pleskdistup/actions/systemd.py
@@ -82,7 +82,7 @@ class DisablePleskRelatedServicesDuringUpgrade(action.ActiveAction):
     oneshot_services: typing.List[str]
 
     def __init__(self):
-        self.name = "rule plesk services"
+        self.name = "disable plesk related services"
         plesk_known_systemd_services = [
             "crond.service",
             "dovecot.service",
@@ -102,7 +102,9 @@ class DisablePleskRelatedServicesDuringUpgrade(action.ActiveAction):
             "sw-cp-server.service",
             "sw-engine.service",
         ]
-        self.plesk_systemd_services = [service for service in plesk_known_systemd_services if systemd.is_service_exists(service)]
+        self.plesk_systemd_services = [
+            service for service in plesk_known_systemd_services if systemd.is_service_startable(service)
+        ]
 
         # Oneshot services are special, so they shouldn't be started on revert or after conversion, just enabled
         self.oneshot_services = [
@@ -111,9 +113,9 @@ class DisablePleskRelatedServicesDuringUpgrade(action.ActiveAction):
 
         # We don't remove postfix service when remove it during qmail installation
         # so we should choose the right smtp service, otherwise they will conflict
-        if systemd.is_service_exists("qmail.service"):
+        if systemd.is_service_startable("qmail.service"):
             self.plesk_systemd_services.append("qmail.service")
-        elif systemd.is_service_exists("postfix.service"):
+        elif systemd.is_service_startable("postfix.service"):
             self.plesk_systemd_services.append("postfix.service")
 
     def _prepare_action(self) -> action.ActionResult:
--- a/pleskdistup/common/src/systemd.py
+++ b/pleskdistup/common/src/systemd.py
@@ -3,7 +3,7 @@ import os
 import typing
 import subprocess
 
-from . import dist, util
+from . import dist, log, util
 
 SYSTEMCTL_BIN_PATH = "/usr/bin/systemctl"
 if dist.get_distro().deb_based:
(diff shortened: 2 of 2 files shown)


pleskdistup/actions/systemd.py
```diff
@@ -102,7 +102,9 @@ def __init__(self):
             "sw-cp-server.service",
             "sw-engine.service",
         ]
-        self.plesk_systemd_services = [service for service in plesk_known_systemd_services if systemd.is_service_exists(service)]
+        self.plesk_systemd_services = [
+            service for service in plesk_known_systemd_services if systemd.is_service_can_be_started(service)
```
kpushkaryov, line 106: Propose to rename the function to `is_service_startable()`.

pleskdistup/common/src/systemd.py
```diff
@@ -24,6 +24,35 @@ def is_service_active(service: str):
     return res.returncode == 0
 
 
+def get_required_services(service: str) -> typing.List[str]:
+    res = subprocess.run(
+        [SYSTEMCTL_BIN_PATH, 'cat', service],
```
kpushkaryov, line 29: I doubt that is reliable. For example, one can create dependencies by putting symlinks into a `<service>.requires` directory instead of declaring them in backing files. Also, there's the reverse form — `RequiredBy` and alternatives, like `Requisite`.
Likely, you should use `systemctl show --property` instead.

pleskdistup/common/src/systemd.py
```diff
@@ -24,6 +24,35 @@ def is_service_active(service: str):
...
+
+
+def is_service_can_be_started(service: str) -> bool:
+    if not is_service_exists(service):
+        return False
+
+    required_services = get_required_services(service)
+    for required_service in required_services:
+        if not is_service_exists(required_service):
+            log.debug("Service '{}' can't be started because required service '{}' doesn't exist".format(service, required_service))
```
kpushkaryov, line 51: Please use an f-string.

kpushkaryov (changes requested):
What's the use case for these changes? Why do we care if required services exist?

The check seems to be very limited. It checks only first-order required services existence, but doesn't check if they are startable. Also, a service could be unstartable even if required services exist.

SandakovMM:
> What's the use case for these changes? Why do we care if required services exist?

Some requirements could not be met, even for Plesk services. We identified an issue with the plesk-ext-monitoring-hcd service, which cannot be started without the Grafana extension installed. Therefore, we need to verify if the Grafana service at least exists. Ideally can be started.

> The check seems to be very limited. It checks only first-order required services existence, but doesn't check if they are startable. Also, a service could be unstartable even if required services exist.

Yes, checking if all of the services could be started might be a good idea.

pleskdistup/actions/systemd.py
kpushkaryov, line 1: Please also fix `self.name = "rule plesk services"`, as it's not descriptive enough and the word "rule" seems to be not suitable.

pleskdistup/actions/systemd.py
```diff
@@ -102,7 +102,9 @@ def __init__(self):
             "sw-cp-server.service",
             "sw-engine.service",
         ]
-        self.plesk_systemd_services = [service for service in plesk_known_systemd_services if systemd.is_service_exists(service)]
+        self.plesk_systemd_services = [
+            service for service in plesk_known_systemd_services if systemd.is_service_startable(service)
```
kpushkaryov, line 106: Apparently, already disabled services should be skipped too.

pleskdistup/common/src/systemd.py
```diff
@@ -24,6 +24,40 @@ def is_service_active(service: str):
...
+        stderr=subprocess.PIPE,
+        check=True,
+        universal_newlines=True
+    )
+
+    required_services = [service for service in res.stdout.split('s=')[1].split() if '.service' in service]
+    return required_services
+
+
+def is_service_startable(service: str, already_checked: typing.Set[str] = None) -> bool:
```
kpushkaryov, line 40: I suppose, masked services shall also be considered unstartable.

kpushkaryov:
> Some requirements could not be met, even for Plesk services. We identified an issue with the plesk-ext-monitoring-hcd service, which cannot be started without the Grafana extension installed.

And what happens then?

kpushkaryov (changes requested):

SandakovMM:
> And what happens then?

Because we are using mass operations for systemctl, it will be seen as failed. At least on CentOS 7 and possibly on older versions of Ubuntu as well, but I am not certain. As a result, we will not be able to continue process (reversion process in this case).

pleskdistup/actions/systemd.py
SandakovMM: We can skip disabled services during the prepare stage, but we shouldn't skip them during the finishing stage. If we do, they will remain disabled and stopped, which will make Plesk unavailable. Since systemd is fine with disabling already disabled services, I suggest leaving it as it is.
I can see the issue when we enable disabled services during conversion/upgrade. However, it doesn't seem like a big problem at the moment. Additionally, I don't want to make more changes in the scope of this PR. We could create an issue to address the problem

kpushkaryov (approved):
OK, but I'm concerned about inadvertently enabling services which were disabled on purpose. If someone disabled the service before us, why should we enable it?

Please rebase before merge.

SandakovMM:
> If someone disabled the service before us, why should we enable it?

Ideally, we wouldn't do this, but in this situation, we have to store a list of the services we have been working with. I suggest creating an issue to address this.

PR #22, opened 2024-03-28
Fix fail on checking if disabled service is masked

SandakovMM:


pleskdistup/common/src/systemd.py
```diff
@@ -42,9 +42,12 @@ def is_service_masked(service: str) -> bool:
         [SYSTEMCTL_BIN_PATH, 'is-enabled', service],
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
-        check=True,
         universal_newlines=True
     )
+    # Because disabled service is not masked
+    if res.returncode != 0:
```
kpushkaryov, line 48: Won't `systemctl is-enabled` always return != 0 for masked services?

kpushkaryov (changes requested):

pleskdistup/common/src/systemd.py
SandakovMM: Yep, so we shouldn't check the returncode at all

kpushkaryov (approved):
Please rebase before merge.

PR #27, opened 2024-04-11
Don't skip failed actions on preparation stage

SandakovMM:
We need to make all required actions on the preparation stage, so if some action has been failed, we should re-perform it

### DIFF
--- a/pleskdistup/common/src/action.py
+++ b/pleskdistup/common/src/action.py
@@ -359,13 +359,16 @@ class PrepareActionsFlow(ActiveFlow):
         return self.stages
 
     def _is_action_required(self, stage: str, action: ActiveAction) -> bool:
-        # Don't repeat already performed actions (in case of process restart)
+        # Don't repeat already performed and succeeded actions (in case of process restart)
+        # If an action has skipped, we should recheck if it could be skipped again
+        # If an action has failed, we should perform it again to make sure conversion/distupgrade
+        # will be performed correctly
         for stored_action in self.actions_data["actions"]:
             if (
                 stored_action["stage"] == stage
                 and stored_action["name"] == action.name
             ):
-                if stored_action["state"] in (ActionState.SUCCESS, ActionState.FAILED):
+                if stored_action["state"] == ActionState.SUCCESS:
                     return False
 
         return action.is_required()
--- a/pleskdistup/common/tests/actionstests.py
+++ b/pleskdistup/common/tests/actionstests.py
@@ -1,4 +1,5 @@
 # Copyright 2023-2024. WebPros International GmbH. All rights reserved.
+import json
 import os
 from unittest import mock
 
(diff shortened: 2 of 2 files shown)


pleskdistup/common/tests/actionstests.py
```diff
@@ -102,20 +117,53 @@ def test_skip_action(self):
...
+            actions_data_file.write('{ "actions": [ { "stage": "test_recheck_skip_based_on_saved_state", "name" : "saved", "state" : "skip"}] }')
+
+        with PrepareActionsFlowForTests({"test_recheck_skip_based_on_saved_state": [simple_action, saved_action]}) as flow:
+            flow.validate_actions()
+            flow.pass_actions()
+
+        simple_action._prepare_action.assert_called_once()
+        saved_action._prepare_action.assert_called_once()
+
+    def test_skip_failed_saved_state(self):
```
kpushkaryov, line 152: These 3 tests are almost identical. Could they be converted into invocations of a single function with different parameters?

pleskdistup/common/src/action.py
```diff
@@ -359,13 +359,16 @@ def _get_flow(self) -> typing.Dict[str, typing.List[ActiveAction]]:
         return self.stages
 
     def _is_action_required(self, stage: str, action: ActiveAction) -> bool:
-        # Don't repeat already performed actions (in case of process restart)
+        # Don't repeat already performed and succeed actions (in case of process restart)
+        # If an action was skipped, we should recheck if it could be skipped again
+        # If an action was failed, we should perform it again to make sure conversion/distupgrade
```
kpushkaryov, line 364: was failed -> has failed

pleskdistup/common/src/action.py
```diff
@@ -359,13 +359,16 @@ def _get_flow(self) -> typing.Dict[str, typing.List[ActiveAction]]:
         return self.stages
 
     def _is_action_required(self, stage: str, action: ActiveAction) -> bool:
-        # Don't repeat already performed actions (in case of process restart)
+        # Don't repeat already performed and succeed actions (in case of process restart)
```
kpushkaryov, line 362: succeed -> succeeded

kpushkaryov (changes requested):

pleskdistup/common/tests/actionstests.py
SandakovMM: Not sure it will simplify things, but ok let's do it

pleskdistup/common/tests/actionstests.py
```diff
@@ -44,6 +45,63 @@ def __init__(self, stages):
...
+    def _revert_action(self) -> action.ActionResult:
+        return action.ActionResult()
+
+
+def check_saved_state(phase, stage, saved_action_state, should_be_called):
+    simple_action = SimpleAction()
+    saved_action = SavedAction()
+
+    if phase == "prepare":
+        simple_action._prepare_action = mock.Mock(return_value=action.ActionResult())
```
kpushkaryov, line 68: I suggest to use `setattr()` and `getattr()` to avoid duplicating these assignments.

kpushkaryov (changes requested):

kpushkaryov (approved):
Please rebase before merge.

PR #32, opened 2024-05-02
Added a function to check if the repository endpoint is none

SandakovMM:
Leapp is not able to handle these types of repositories, so we should refrain from adding them.

### DIFF
--- a/pleskdistup/common/src/rpm.py
+++ b/pleskdistup/common/src/rpm.py
@@ -216,3 +216,17 @@ def upgrade_packages(pkgs: typing.Optional[typing.List[str]] = None) -> None:
 
 def autoremove_outdated_packages() -> None:
     util.logged_check_call(["/usr/bin/yum", "autoremove", "-y"])
+
+
+def repository_has_none_link(
+    id: typing.Optional[str],
+    name: typing.Optional[str],
+    url: typing.Optional[str],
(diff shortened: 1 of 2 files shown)


pleskdistup/common/src/rpm.py
```diff
@@ -216,3 +216,17 @@ def upgrade_packages(pkgs: typing.Optional[typing.List[str]] = None) -> None:
 
 def autoremove_outdated_packages() -> None:
     util.logged_check_call(["/usr/bin/yum", "autoremove", "-y"])
+
+
+def is_repository_has_none_link(
```
kpushkaryov, line 221: "is... has" sounds weird. Propose to drop "is".

pleskdistup/common/tests/rpmtests.py
```diff
@@ -377,3 +377,23 @@ def test_handle_whole_directory(self):
             self.assertEqual(open(full_filepath).read(), content)
 
         shutil.rmtree(self.test_dir)
+
+
+class isRepositoryHasNoneLinkTest(unittest.TestCase):
```
kpushkaryov, line 382: See above.

kpushkaryov (changes requested):

kpushkaryov (approved):

PR #33, opened 2024-05-10
Use utf-8 encoding on file IO to avoid problems on host with changed default encoding

SandakovMM:


kpushkaryov (changes requested):
Hardcoding `encoding=` everywhere violates the DRY principle a lot, cluttering the code and introducing the possibility to forget adding that later. If we need this change, global default switch seems preferable.

But I'm not sure that we need it. Please explain the problem.

SandakovMM:
> But I'm not sure that we need it. Please explain the problem.

We have a bug related to the problem in our internal broad in jira with the STR.
In short, when customer has unusual locale and utf-8 ecoded files, we got problems o parsing.
Several customers have already encountered this issue, and I anticipate more as we have similar problems in Plesk itself

SandakovMM:
> Hardcoding encoding= everywhere violates the DRY principle a lot, cluttering the code and introducing the possibility to forget adding that later. If we need this change, global default switch seems preferable.

It would be beneficial to set it globally, but I haven't found a way to do so. I've attempted:
1. using sys.setdefaultencoding, but it's not available on Python 3
2. resetting environment variables LANG and LC_ALL from inside the tool, but it doesn't seem to solve the issue
I would appreciate it if you could share another way to achieve this

kpushkaryov:
> > Hardcoding encoding= everywhere violates the DRY principle a lot, cluttering the code and introducing the possibility to forget adding that later. If we need this change, global default switch seems preferable.
> 
> It would be beneficial to set it globally, but I haven't found a way to do so. I've attempted:
> 
>     1. using sys.setdefaultencoding, but it's not available on Python 3
> 
>     2. resetting environment variables LANG and LC_ALL from inside the tool, but it doesn't seem to solve the issue
>        I would appreciate it if you could share another way to achieve this

I checked Python 3.6 source code: `open()` calls end up in `TextIOWrapper`, which [uses](https://github.com/python/cpython/blob/5c4568a05a0a62b5947c55f68f9f2ecfb90a4f12/Lib/_pyio.py#L1929C32-L1930C1) `locale.getpreferredencoding(False)` to determine default file encoding, which in turn boils down to `nl_langinfo(CODESET)` call in C API. Thus, you can change the default encoding by `locale.setlocale(locale.LC_ALL, 'en_US.UTF-8')`.

SandakovMM:
> Thus, you can change the default encoding by locale.setlocale(locale.LC_CTYPE, 'en_US.UTF-8').

It does not help. Example script:
```py
#!/usr/bin/python3

import locale

print(locale.getpreferredencoding(False))
locale.setlocale(locale.LC_CTYPE, 'en_US.UTF-8')
print("alalal")
print("ü")

```

SandakovMM:
> It does not help.

I mean stdout obviously, because it was opened before we change locale. I don't want to change default stdout actually. However we could just fix `StdoutWriter` here. Looks promising, let me try it

kpushkaryov:
> It does not help. Example script:

This PR is about file IO, isn't it? Stdout encoding is a different matter.

pleskdistup/common/src/writers.py
```diff
@@ -18,9 +18,15 @@ def __exit__(self, *args):
 
 
 class StdoutWriter(Writer):
+    def __init__(self):
+        self.out = open(1, 'w', closefd=False)
```
kpushkaryov, line 22: Are we actually sending non-ASCII characters to StdoutWriter?

pleskdistup/main.py
```diff
@@ -356,13 +357,20 @@ def main():
         "-v", "--version", action="store_true",
         help="show the version of this utility."
     )
+    parser.add_argument("--locale", default="en_US.UTF-8", help="set the locale for the utility")
```
kpushkaryov, line 360: We need to carry this option to resume stage, otherwise the utility can fail after reboot.

kpushkaryov (changes requested):

SandakovMM:
> This PR is about file IO, isn't it?

Both. Because we have the problem for stdout as well.

pleskdistup/common/src/writers.py
SandakovMM: Yes, for example, this can happen when we describe the pre-checker results, also known as "required conditions."

kpushkaryov: Could you provide an example?

SandakovMM: Sure. This is the STR from the bug we are fixing here:
1. Add utf-8 symbol in the /etc/mail/spamassassin/init.pre. For example 'ü'
2. set locale with environment variables:
```
> LC_ALL=C
> LANG=C
```
3. Start centos2alma

Currently, the bug causes a failure while reading the file on centos2alma 1.3.0. Even if we fix the default encoding by using setlocale, it will still fail when writing the spamassassin plugin name to stdout for the same reason.

pleskdistup/main.py
```diff
@@ -356,13 +357,20 @@ def main():
...
     if getattr(options, "help", False):
         parser.print_help()
     else:
         options.help = False
 
+    # Configure locale to avoid problems on systems where LANG or LC_CTYPE changed,
+    # while files on the system still has utf-8 encoding
+    # We should do it before initializing logger to make sure utf-8 symbols will be
+    # written correctly to the log file.
+    locale.setlocale(locale.LC_CTYPE, options.locale)
```
kpushkaryov, line 372: I think, we shouldn't call `setlocale()` if `--locale` isn't set, otherwise we would be always overriding user system settings.

kpushkaryov (changes requested):
What's the root cause of the encoding discrepancy we're trying to solve here? Why files contain UTF-8, while system locale is not UTF-8?

Fixing the system is more robust approach than trying to workaround its inconsistent state inside dist-upgrader. I'm afraid we can run into further problems doing that, e.g. we set dist-upgrader own encoding, but what encoding will be used by utilities it calls? Will many operations carried out during Plesk and system upgrade be carried out correcty in such environment?

pleskdistup/main.py
SandakovMM: Actually I can't see any reason to have `--local` then, we could just ask a customer simply use environment variables.

pleskdistup/main.py
```diff
@@ -356,13 +357,21 @@ def main():
         "-v", "--version", action="store_true",
         help="show the version of this utility."
     )
+    parser.add_argument("--locale", default=None, help=argparse.SUPPRESS)
```
kpushkaryov, line 360: Please maintain alphabetical order of options.

kpushkaryov (changes requested):

kpushkaryov (approved):
Please rebase before merge.

PR #35, opened 2024-05-14
Handle UnicodeDecodeError carefully to propose customer a way to workaround the problem

SandakovMM:


SandakovMM:
This is another method to prevent the "ascii codec encoding problem." It may be slightly inconvenient for customers, but I believe it could be safer for us because we are not relying on any unchangeable defaults.

pleskdistup/main.py
```diff
@@ -246,6 +253,13 @@ def do_convert(
                 elif options.phase is Phase.FINISH:
                     print(messages.FINISH_RESTART_MESSAGE, end='')
                 systemd.do_reboot()
+        except UnicodeDecodeError as ex:
+            if os.environ['LANG'] != "en_US.UTF-8" or os.environ["LC_ALL"]!= "en_US.UTF-8":
+                printerr(f"One of your file encoding doesn't fit with locale set through environment. Call the tool again with 'LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 ./centos2alma'")
+            else:
+                ex_info = traceback.format_exc()
+                printerr(f"Preparation checks failed: {ex}\n{ex_info}")
+            return False
```
kpushkaryov, line 262: The function returns `int`.

pleskdistup/main.py
```diff
@@ -246,6 +253,13 @@ def do_convert(
                 elif options.phase is Phase.FINISH:
                     print(messages.FINISH_RESTART_MESSAGE, end='')
                 systemd.do_reboot()
+        except UnicodeDecodeError as ex:
+            if os.environ['LANG'] != "en_US.UTF-8" or os.environ["LC_ALL"]!= "en_US.UTF-8":
+                printerr(f"One of your file encoding doesn't fit with locale set through environment. Call the tool again with 'LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 ./centos2alma'")
+            else:
+                ex_info = traceback.format_exc()
+                printerr(f"Preparation checks failed: {ex}\n{ex_info}")
```
kpushkaryov, line 261: Is it actually preparation checks failure?

pleskdistup/main.py
kpushkaryov, line 262: Why `handle_error()` isn't used?

kpushkaryov (changes requested):

pleskdistup/common/src/action.py
```diff
@@ -467,6 +472,8 @@ def make_checks(self) -> typing.List[str]:
                     failed_checks_msgs.append(
                         f"Required pre-conversion condition {check.name!r} not met:\n\t{check.description}\n"
                     )
+            except UnicodeDecodeError as e:
```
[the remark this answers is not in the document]
SandakovMM: Yep, because I want to be able to identify the exact UnicodeDecodeError exception at the next level.
If I understand correctly, the next clause will replace the type with RuntimeError, concealing the original error inside.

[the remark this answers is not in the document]
kpushkaryov: Losing `check.name` isn't good. Instead, for example, you can check `RuntimeError.__cause__`.

pleskdistup/main.py
```diff
@@ -246,6 +254,12 @@ def do_convert(
                 elif options.phase is Phase.FINISH:
                     print(messages.FINISH_RESTART_MESSAGE, end='')
                 systemd.do_reboot()
+        except UnicodeDecodeError as e:
+            if locale.getpreferredencoding(False).lower() != "utf-8":
+                printerr(f"One of your file encoding doesn't fit with locale set through environment. Call the tool again with 'LC_CTYPE=en_US.UTF-8 {os.path.abspath(sys.argv[0])}'")
```
kpushkaryov, line 259: Why this case isn't handled by `handle_error()` too?

kpushkaryov (changes requested):
This PR and #33 solve the same problem. Which do you want to merge?

SandakovMM:
> This PR and https://github.com/plesk/dist-upgrader/pull/33 solve the same problem. Which do you want to merge?

I want to merge #33. After that, we can update the error messages in this PR to emphasize the `--local` option.

kpushkaryov (approved):
Please rebase before merge.

PR #37, opened 2024-05-22
Show whitch exactly check action is bad encoded

SandakovMM:
Otherwise it's preatty hard to understand where the customer should look for a 'bad' encoded file.

### DIFF
--- a/pleskdistup/common/src/log.py
+++ b/pleskdistup/common/src/log.py
@@ -1,4 +1,5 @@
 # Copyright 2023-2024. WebPros International GmbH. All rights reserved.
+import locale
 import logging
 
 import typing
@@ -9,10 +10,20 @@ class logger():
 
     is_streams_enabled = False
     streams_logger = logging.getLogger("distupgrade_streams")
+    encoding: str = locale.getpreferredencoding()
+
+    @staticmethod
+    def _re_decode_message(message: str) -> str:
+        return message.encode(logger.encoding, errors='backslashreplace').decode(logger.encoding, errors='backslashreplace')
 
     @staticmethod
     def init_logger(logfiles: typing.List[str], streams: typing.List[typing.Any],
-                    console: bool = False, loglevel: int = logging.INFO) -> None:
+                    console: bool = False, loglevel: int = logging.INFO, encoding: str = None) -> None:
+        if encoding is None:
+            logger.encoding = locale.getpreferredencoding()
+        else:
+            logger.encoding = encoding
+
         logger.files_logger.setLevel(loglevel)
         logger.streams_logger.setLevel(loglevel)
 
@@ -39,13 +50,14 @@ class logger():
 
     @staticmethod
     def reinit_logger(logfiles: typing.List[str], streams: typing.List[typing.Any],
-                      console: bool = False, loglevel: int = logging.INFO) -> None:
+                      console: bool = False, loglevel: int = logging.INFO, encoding: str = None) -> None:
         logger.files_logger = logging.getLogger("distupgrade_files")
         logger.streams_logger = logging.getLogger("distupgrade_streams")
-        logger.init_logger(logfiles, streams, console, loglevel)
+        logger.init_logger(logfiles, streams, console, loglevel, encoding=encoding)
 
     @staticmethod
     def debug(msg: str, to_file: bool = True, to_stream: bool = True) -> None:
+        msg = logger._re_decode_message(msg)
         if to_file:
             logger.files_logger.debug(msg)
 
@@ -54,6 +66,7 @@ class logger():
 
     @staticmethod
     def info(msg: str, to_file: bool = True, to_stream: bool = True) -> None:
+        msg = logger._re_decode_message(msg)
         if to_file:
             logger.files_logger.info(msg)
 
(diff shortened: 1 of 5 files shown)


pleskdistup/main.py
```diff
@@ -101,14 +101,19 @@ def required_conditions_satisfied(upgrader: DistUpgrader, options: typing.Any, p
...
             writer.write("\r")
             for check in failed_checks:
-                writer.write(check)
-                log.err(check)
+                try:
+                    writer.write("Required pre-conversion condition {check.name!r} not met:\n\t{check.description}\n")
+                    log.err(check)
+                except (UnicodeEncodeError, UnicodeDecodeError) as ex:
+                    bad_encoded_check = check.name
+                    raise RuntimeError(f"Encoding error on writing problem description") from ex
```
kpushkaryov, line 116: Pointless f-string.

pleskdistup/main.py
```diff
@@ -101,14 +101,19 @@ def required_conditions_satisfied(upgrader: DistUpgrader, options: typing.Any, p
...
         with action.CheckFlow(checks) as check_flow, writers.StdoutWriter() as writer:
             writer.write("Doing preparation checks...\n")
             check_flow.validate_actions()
             failed_checks = check_flow.make_checks()
             writer.write("\r")
             for check in failed_checks:
-                writer.write(check)
-                log.err(check)
+                try:
+                    writer.write("Required pre-conversion condition {check.name!r} not met:\n\t{check.description}\n")
```
kpushkaryov, line 112: Omitted f-string.

kpushkaryov (changes requested):
Typo in the commit message: whitch -> which.

This output errors processing is getting messy. Could we instead use an alternative [error handler](https://docs.python.org/3/library/functions.html#open), like `replace`?

SandakovMM:
> This output errors processing is getting messy. Could we instead use an alternative [error handler](https://docs.python.org/3/library/functions.html#open), like replace?

We could use 'replace' for pre-checks. But then we would be hiding the inconsistency in locale from the user.
I would like to clearly identify the problem.

kpushkaryov:
> We could use 'replace' for pre-checks. But then we would be hiding the inconsistency in locale from the user. I would like to clearly identify the problem.

I mean, use `replace` or its analog for terminal and log output. Otherwise, we can't output the problematic message itself, even if it's 99% ASCII, thus reducing diagnostic potential.

SandakovMM:
> I mean, use replace or its analog for terminal and log output

Yep, but unfortunately there is no errors argument for logging.FileHandler in python 3.6. So we can't simple change error handling for logs.

pleskdistup/common/src/writers.py
```diff
@@ -29,6 +29,28 @@ def __exit__(self, *args):
...
+
+    def write(self, message: str) -> None:
+        self.out.write(message)
+        self.out.flush()
+
+        recoded_message = codecs.encode(message, self.out.encoding, errors='replace').decode(self.out.encoding)
+        if recoded_message != message:
+            self.encoding_errors_met = True
+
+    def is_there_encoding_errors(self) -> bool:
```
kpushkaryov, line 47: `is_there_encoding_errors` is weird, I suggest `has_encoding_errors`.

pleskdistup/common/src/log.py
```diff
@@ -9,10 +10,20 @@ class logger():
 
     is_streams_enabled = False
     streams_logger = logging.getLogger("distupgrade_streams")
+    encoding: str = locale.getpreferredencoding()
+
+    @staticmethod
+    def _re_decode_message(message: str) -> str:
+        return str(message).encode(logger.encoding, errors='backslashreplace').decode(logger.encoding, errors='backslashreplace')
```
kpushkaryov, line 17: According to annotation, `message` should be also `str`, why `str()` then?

pleskdistup/main.py
```diff
@@ -101,19 +101,19 @@ def required_conditions_satisfied(upgrader: DistUpgrader, options: typing.Any, p
...
-                    log.err(check)
-                except (UnicodeEncodeError, UnicodeDecodeError) as ex:
-                    bad_encoded_check = check.name
-                    raise RuntimeError(f"Encoding error on writing problem description") from ex
+                writer.write(f"Required pre-conversion condition {check.name!r} not met:\n\t{check.description}\n")
+                log.err(check)
+
+            if writer.is_there_encoding_errors():
+                printerr(messages.ENCODING_INCONSISTENCY_ERROR_MESSAGE)
+                printerr("The encoding problem occurred during one of the previous check. The badly encoded symbol was replaced with a backslash Unicode representation")
```
kpushkaryov, line 115: check -> checks

kpushkaryov (changes requested):

pleskdistup/common/src/log.py
SandakovMM: We sometimes send objects into log functions, like in this [example] (https://github.com/plesk/dist-upgrader/blob/main/pleskdistup/main.py#L111) , where a failed pre-check is logged to the error log.
Likely we could just change type annotation to "something convertible to string"

SandakovMM: Ah, wait a second, the place was actually broken by this change, we just need to fix it

kpushkaryov (dismissed):
Please rebase before merge. Don't forget to fix the typo in the commit message.

kpushkaryov (approved):

PR #38, opened 2024-05-24
Enable mypy linter checks on pull requests

SandakovMM:


pleskdistup/common/src/dist.py
```diff
@@ -1,4 +1,5 @@
 # Copyright 2023-2024. WebPros International GmbH. All rights reserved.
+from functools import lru_cache
```
kpushkaryov, line 2: Keep `from` below `import`.

kpushkaryov (changes requested):
Typo in commit message: "standart python" -> "standard Python".

PR #39, opened 2024-05-27
TuxCare ELS support and general fixes

vizovitin:
This is a pre-requisite for other TuxCare ELS support PRs. PPP-64910

### DIFF
--- a/pleskdistup/actions/extensions.py
+++ b/pleskdistup/actions/extensions.py
@@ -4,7 +4,7 @@ import pwd
 import shutil
 import typing
 
-from pleskdistup.common import action, dist, files, log, systemd, util
+from pleskdistup.common import action, dist, files, plesk, log, systemd, util
 
 
 class DisableGrafana(action.ActiveAction):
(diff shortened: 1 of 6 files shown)


pleskdistup/actions/extensions.py
```diff
@@ -133,3 +133,33 @@ def _revert_action(self) -> action.ActionResult:
...
+
+    def __init__(self):
+        self.name = "uninstall tuxcare-els"
+        self.ext_name = "tuxcare-els"
+
+    def _is_required(self) -> bool:
+        return self.ext_name in dict(plesk.list_installed_extensions())
+
+    def _prepare_action(self) -> action.ActionResult:
+        util.logged_check_call(["/usr/sbin/plesk", "bin", "extension", "--uninstall", self.ext_name])
```
SandakovMM, line 151: NIT: could be moved to the plesk module since we already have list_installed_extensions

pleskdistup/actions/extensions.py
```diff
@@ -133,3 +133,33 @@ def _revert_action(self) -> action.ActionResult:
...
+        return action.ActionResult()
+
+    def _post_action(self) -> action.ActionResult:
+        return action.ActionResult()
+
+    def _revert_action(self) -> action.ActionResult:
+        util.logged_check_call(["/usr/sbin/plesk", "bin", "extension", "--install", self.ext_name])
```
SandakovMM, line 158: NIT: could be moved to the plesk module as well

pleskdistup/actions/extensions.py
```diff
@@ -133,3 +133,33 @@ def _revert_action(self) -> action.ActionResult:
...
+
+
+class UninstallTuxcareEls(action.ActiveAction):
+    """ TuxCare ELS extension is installed on EoL OSes and may enable repositories incompatible with
+        the target OS version. Uninstalling also removes the repositories.
+    """
+
+    def __init__(self):
+        self.name = "uninstall tuxcare-els"
+        self.ext_name = "tuxcare-els"
```
kpushkaryov, line 145: Please declare `ext_name`'s type on class level for type checking.

kpushkaryov (changes requested):

pleskdistup/actions/extensions.py
vizovitin: Type checking is not set up properly anyway, so this has no bearing on the checks result or code correctness.
Also mypy is relatively smart and explicit type annotations for class members are often excessive (as long as the constructor is treated as typed, which is not the case here and in majority of other existing code here).
I'll fix it, but I see no current value in this.

vizovitin:
@SandakovMM , @kpushkaryov , any other comments?

kpushkaryov (approved):

SandakovMM (approved):

PR #40, opened 2024-05-31
List of supported php

SandakovMM:


pleskdistup/actions/common_checks.py
```diff
@@ -98,150 +98,229 @@ def _do_check(self) -> bool:
...
-        )
+        self.description = self._format_description(installed_unsupported_php)
 
-        log.debug("Outdated PHP versions found")
         return False
 
 
-class AssertMinPhpVersionUsedByWebsites(action.CheckAction):
-    min_version: version.PHPVersion
+class AssertMinPhpVersionInstalled(AssertInstalledPhpVersionsInList):
```
kpushkaryov, line 138: `AssertMinPhpVersionInstalled` shouldn't be a subclass of `AssertInstalledPhpVersionsInList`, because the former asserts $version >= min$ and the latter $version \in Allowed$. Thus, `AssertMinPhpVersionInstalled` can't fulfill the role of `AssertInstalledPhpVersionsInList`, which violates the Liskov substitution principle.

Also now `AssertMinPhpVersionInstalled` would fail encountering a future unknown version of PHP.

pleskdistup/actions/common_checks.py
```diff
@@ -98,150 +98,229 @@ def _do_check(self) -> bool:
...
+    def __init__(
+        self,
+        min_version: str,
+    ):
+        self.name = "check for outdated PHP versions"
```
kpushkaryov, line 143: Base class constructor not called here and in the following classes.

pleskdistup/actions/common_checks.py
```diff
@@ -98,150 +98,229 @@ def _do_check(self) -> bool:
...
             log.err("Unable to get domains list from plesk database!")
             raise RuntimeError("Unable to get domains list from plesk database!") from ex
 
         return False
 
 
-class AssertMinPhpVersionUsedByCron(action.CheckAction):
-    min_version: version.PHPVersion
-    optional: bool
+class AssertMinPhpVersionUsedByWebsites(AssertPhpVersionsUsedByWebsitesInList):
```
kpushkaryov, line 214: Substitution principle violation, see `AssertMinPhpVersionInstalled`.

pleskdistup/actions/common_checks.py
```diff
@@ -98,150 +98,229 @@ def _do_check(self) -> bool:
...
+
+    def _format_description(self, installed_unsupported_php: typing.List[str]) -> str:
+        return self.description.format(
+            versions=", ".join(installed_unsupported_php),
+            remove_arg=" ".join([php_version.replace(" ", "") for php_version in installed_unsupported_php]).lower()
+        )
+
+
+class AssertPhpVersionsUsedByWebsitesInList(action.CheckAction):
+    target_versions: typing.List[version.PHPVersion]
```
kpushkaryov, line 159: The same as for `target_versions` above.

pleskdistup/actions/common_checks.py
```diff
@@ -98,150 +98,229 @@ def _do_check(self) -> bool:
...
+
+    def _format_description(self, extraneous_php_domains: typing.List[str]) -> str:
+        return self.description.format(
+                modern=self.min_version,
+                domains=extraneous_php_domains
+            )
+
+
+class AssertPhpVersionsUsedByCronInList(action.CheckAction):
+    target_versions: typing.List[version.PHPVersion]
```
kpushkaryov, line 242: See above.

pleskdistup/common/src/php.py
```diff
@@ -22,3 +22,11 @@ def get_php_handlers(php_versions: typing.List[version.PHPVersion]) -> typing.Li
 
 def get_outdated_php_handlers(first_modern: version.PHPVersion) -> typing.List[str]:
     return get_php_handlers([php for php in get_known_php_versions() if php < first_modern])
+
+
+def get_php_handlers_by_condition(condition: typing.Callable[[version.PHPVersion], bool]) -> typing.List[str]:
+    return get_php_handlers([php for php in get_known_php_versions() if condition(php)])
+
+
+def get_php_versions_by_condition(condition: typing.Callable[[version.PHPVersion], bool]) -> typing.List[str]:
```
kpushkaryov, line 31: If the condition checks `PHPVersion`, then the function should return `List[PHPVersion]`.

pleskdistup/common/tests/phptests.py
```diff
@@ -76,3 +76,123 @@ def test_no_outdated(self):
             php.get_outdated_php_handlers(version.PHPVersion("5.0")),
             []
         )
+
+
+class TestGetHandlersByCondition(unittest.TestCase):
+    def test_lover_then_version_7_1(self):
```
kpushkaryov, line 82: lover -> lower
then -> than

pleskdistup/common/tests/phptests.py
```diff
@@ -76,3 +76,123 @@ def test_no_outdated(self):
...
+        self.assertEqual(
+            sorted(php.get_php_handlers_by_condition(lambda php: php < version.PHPVersion("7.1"))),
+            [
+                "plesk-php52-cgi",
```
kpushkaryov, line 86: If this test is for `get_php_handlers_by_condition()`, it shouldn't contain a fixed list of handlers. Otherwise, it becomes a test for combination of  `get_php_handlers_by_condition()`, `get_known_php_versions()` and `get_php_handlers()`, which isn't a unit test.

pleskdistup/common/tests/phptests.py
```diff
@@ -76,3 +76,123 @@ def test_no_outdated(self):
...
+                "plesk-php70-fpm-dedicated",
+            ]
+        )
+
+    def test_not_in_list(self):
+        target_versions = [php for php in php.get_known_php_versions() if php >= version.PHPVersion("7.1")]
+        self.assertEqual(
+            sorted(php.get_php_handlers_by_condition(lambda php: php not in target_versions)),
+            [
+                "plesk-php52-cgi",
```
kpushkaryov, line 118: Duplicated list.

pleskdistup/common/tests/phptests.py
```diff
@@ -76,3 +76,123 @@ def test_no_outdated(self):
             php.get_outdated_php_handlers(version.PHPVersion("5.0")),
             []
         )
+
+
+class TestGetHandlersByCondition(unittest.TestCase):
+    def test_lover_then_version_7_1(self):
+        self.assertEqual(
```
kpushkaryov, line 83: [assertListEqual(a, b)](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual)

kpushkaryov (changes requested):
Also typos in commit messages:
- retrive -> retrieve
- possinble -> possible
- Added -> Add

pleskdistup/actions/common_checks.py
SandakovMM: Let's just use condition check as the parent of both

pleskdistup/common/tests/phptests.py
SandakovMM: Yes, it's expected. I'm trying to see how the entire module works here. For example, if the list of known PHP versions changes, I want to make sure the test detects it.
Otherwise, I would only be checking how one line with a condition call works, which doesn't make much sense to me.

pleskdistup/actions/common_checks.py
```diff
@@ -98,150 +98,306 @@ def _do_check(self) -> bool:
...
+            formatter=lambda installed_unsupported_php: self._format_description(installed_unsupported_php),
+            condition=lambda php: php >= self.min_version,
+            optional=optional
+        )
+
+    def _format_description(self, extraneous_php_domains: typing.List[str]) -> str:
+        return self.description.format(
+                modern=self.min_version,
+                cronjobs=extraneous_php_domains
+            )
```
kpushkaryov, line 398: Too much indent.

kpushkaryov (changes requested):

pleskdistup/actions/common_checks.py
```diff
@@ -98,150 +98,306 @@ def _do_check(self) -> bool:
         return False
 
 
-class AssertMinPhpVersionInstalled(action.CheckAction):
-    min_version: version.PHPVersion
+class AssertInstalledPhpVersionsInCondition(action.CheckAction):
```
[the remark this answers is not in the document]
SandakovMM: Maybe we could use 'by' here

kpushkaryov (approved):
Please rebase before merge.
Don't forget to [fix commit messages](https://github.com/plesk/dist-upgrader/pull/40#pullrequestreview-2092773962).

PR #42, opened 2024-06-04
Added an action to verify if the LiteSpeed extension is installed

SandakovMM:
The extension is not yet supported in the centos2alma conversion

### DIFF
--- a/pleskdistup/actions/extensions.py
+++ b/pleskdistup/actions/extensions.py
@@ -164,3 +164,78 @@ class UninstallTuxcareEls(action.ActiveAction):
 
     def estimate_revert_time(self) -> int:
         return 20
+
+
+class AssertPleskExtensions(action.CheckAction):
+    installed: typing.Set[str]
+    not_installed: typing.Set[str]
+    installed_description: str
+    not_installed_description: str
+    installed_violations: typing.Set[str]
+    not_installed_violations: typing.Set[str]
+    _name: str
+
+    def __init__(
+        self,
+        installed: typing.Optional[typing.Iterable[str]] = None,
+        not_installed: typing.Optional[typing.Iterable[str]] = None,
+        name: str = "check Plesk extensions state:",
+        installed_description: typing.Optional[str] = None,
+        not_installed_description: typing.Optional[str] = None,
+    ):
+        if installed is None and not_installed is None:
+            raise ValueError("'installed' and 'not_installed' can't be both None")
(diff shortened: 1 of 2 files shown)


pleskdistup/actions/extensions.py
```diff
@@ -164,3 +164,12 @@ def estimate_prepare_time(self) -> int:
 
     def estimate_revert_time(self) -> int:
         return 20
+
+
+class AssertLiteSpeedNotInstalled(action.CheckAction):
```
kpushkaryov, line 169: This check is single-use. Propose to replace with general `AssertPleskExtensions` (like existing `AssertPleskComponents`).

pleskdistup/actions/extensions.py
```diff
@@ -164,3 +164,12 @@ def estimate_prepare_time(self) -> int:
...
+    def __init__(self):
+        self.name = "checking if LiteSpeed is installed"
+        self.description = """LiteSpeed extension is installed on the system. To proceed with the upgrade, remove the LiteSpeed extension."""
```
kpushkaryov, line 172: Unneeded triple quotes.

kpushkaryov (changes requested):

pleskdistup/actions/extensions.py
SandakovMM: Looks like an overkill, but whatever

pleskdistup/actions/extensions.py
```diff
@@ -164,3 +164,74 @@ def estimate_prepare_time(self) -> int:
...
+    ):
+        if installed is None and not_installed is None:
+            raise ValueError("'installed' and 'not_installed' can't be both None")
+        self.installed = set(installed) if installed is not None else set()
+        self.not_installed = set(not_installed) if not_installed is not None else set()
+        self._name = name
+        self.installed_violations = set()
+        self.not_installed_violations = set()
+        if installed_description is None:
+            self._installed_description = "Required Plesk extensions are missing. To continue, please install the following extensions:\n\t- {installed_violations}"
```
kpushkaryov, line 194: If `installed_description`, `not_installed_description` aren't `None`, they are not going anywhere. The same bug is in `AssertPleskComponents`.

kpushkaryov (changes requested):

kpushkaryov (approved):
Please rebase before merge.
Don't forget to update commit messages, because the change set is not about LiteSpeed anymore.

PR #46, opened 2024-07-09
Fix mypy configuration

kpushkaryov:
Some files were not checked previously.


.github/workflows/mypy.yaml
```diff
@@ -1,6 +1,6 @@
 name: MyPy Lint
 on:
-  pull_request:
+  push
```
SandakovMM, line 3: I'm not certain if we need to lint every push. For instance, checking seems unnecessary when we push tags.
From what I understand, the `pull_request` trigger will initiate the action each time we push to the branch associated with the pull request.

pleskdistup/common/src/files.py
```diff
@@ -46,9 +46,6 @@ def push_front_strings(filename: str, strings: typing.List[str]) -> None:
 
 
 def rewrite_json_file(filename: str, jobj: typing.Union[dict, typing.List]) -> None:
-    if filename is None or jobj is None:
```
SandakovMM, line 49: It might be better to raise an exception here

pleskdistup/common/src/postgres.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 1999 - 2024. Plesk International GmbH. All rights reserved.
+# Copyright 2023-2024. WebPros International GmbH. All rights reserved.
```
SandakovMM, line 1: I suggest fixing all the copyrights

SandakovMM (changes requested):
:arrow_up:

pleskdistup/common/src/postgres.py
kpushkaryov: Others are fine.

pleskdistup/common/src/files.py
kpushkaryov: > It might be better to raise an exception here

Arguments are statically annotated as not None and actually we don't check arguments dynamically for None anywhere in this file.

SandakovMM (approved):

PR #50, opened 2024-07-11
Move reboot out of lock

SandakovMM:


pleskdistup/convert.py
```diff
@@ -47,12 +47,15 @@ def start_flow(
 # Unfortunately, dataclasses available since Python 3.7 aren't supported
 # on Ubuntu 18 (Python 3.6)
 class ConvertResult:
+    successful: bool
```
kpushkaryov, line 50: Just `success` would be enough, I think.

kpushkaryov (changes requested):

kpushkaryov (approved):
Please rebase before merge.

PR #53, opened 2024-07-22
Verify that the Plesk database is functioning before attempting to collaborate on any Plesk extensions

SandakovMM:


pleskdistup/common/src/plesk.py
```diff
@@ -124,10 +127,16 @@ def list_installed_extensions() -> typing.List[typing.Tuple[str, str]]:
 
 
 def install_extension(name: str) -> None:
+    if not is_plesk_database_ready():
```
vizovitin, line 130: Note that this checks only if the service is up, not that it can service any requests (e.g. authentication works correctly and DB is alive).

pleskdistup/actions/extensions.py
```diff
@@ -146,21 +146,27 @@ def __init__(self) -> None:
         self.ext_name = "tuxcare-els"
 
     def _is_required(self) -> bool:
-        # If database is not ready, we can't be sure about installed extensions
-        if (not plesk.is_plesk_database_ready()):
+        try:
+            return self.ext_name in dict(plesk.list_installed_extensions())
+        except plesk.PleskDatabaseIsDown:
+            # If database is not ready we will not be able to uninstall the extension anyway
```
vizovitin, line 152: At least write a warning. Because this may result in failures down the line.

pleskdistup/actions/extensions.py
```diff
@@ -146,21 +146,27 @@ def __init__(self) -> None:
...
             return False
 
-        return self.ext_name in dict(plesk.list_installed_extensions())
-
     def _prepare_action(self) -> action.ActionResult:
-        plesk.uninstall_extension(self.ext_name)
+        try:
+            plesk.uninstall_extension(self.ext_name)
+        except plesk.PleskDatabaseIsDown:
+            log.warn("Removing TuxCare ELS extension called when database is already down")
```
vizovitin, line 159: This is written like it's intended for a developer. For a user it's unclear what to do with this.

pleskdistup/common/src/plesk.py
vizovitin, line 1: Wouldn't it be better to have some general pre-check instead (or in addition to this)?

vizovitin (changes requested):

pleskdistup/common/src/plesk.py
SandakovMM: The change mainly ensures there are no problems on `revert`, as we know the database service has been stopped on preparation stage.
So far, we haven't had any issues with incorrect authentication or database corruption. If such problems do occur, a thorough investigation would be needed. It's likely to cause the entire conversion process to fail, so skipping some conversion steps doesn't make much sense. So yeah, it's kind of expected for this change,

pleskdistup/actions/extensions.py
SandakovMM: This message is intended for developers. It indicates that the order of actions is probably incorrect or that previous actions have failed. It's unlikely that a user will be able to fix the problem on their own.

pleskdistup/common/src/plesk.py
SandakovMM: We don't perform pre-checks during a revert because, at that stage, we already know everything is broken. Therefore, adding a pre-check for the database doesn't make much sense. Additionally, we can be confident that the database will be stopped before the revert (on preparation stage).
Regarding `UninstallTuxcareEls`, we determine that an action should be reverted based on the actions.json file, which records what actions were taken during the preparation step. The original bug occurred when the actions.json file was manually removed. However, I would like the revert process to work even if the actions.json file is somehow removed.

pleskdistup/common/src/plesk.py
vizovitin: OK

vizovitin (approved):
Please rebase.

PR #67, opened 2024-08-29
Adopt ubuntu php73

SandakovMM:


pleskdistup/common/src/plesk.py
```diff
@@ -237,3 +237,23 @@ def get_from_plesk_database(query: str) -> typing.Optional[typing.List[str]]:
...
     return proc.stdout.splitlines()
+
+
+def get_repository_by_os_from_inf3(inf3_content: typing.Union[ElementTree.Element, str], os: dist.Distro) -> typing.Optional[str]:
+    if isinstance(inf3_content, str):
+        if not inf3_content:
+            return None
+        inf3_content = ElementTree.fromstring(inf3_content)
+
+    for build in inf3_content.findall(".//build"):
```
kpushkaryov, line 248: Won't just "//build" work?

pleskdistup/common/tests/plesktests.py
```diff
@@ -88,3 +88,43 @@ def test_only_plesk_without_release_key(self):
 """
 
         self.assertEqual([], plesk.extract_plesk_versions(data))
+
+
+class TestGetRepositoryByOSFromInf3(unittest.TestCase):
```
kpushkaryov, line 93: OS -> Os

kpushkaryov (changes requested):

pleskdistup/common/src/plesk.py
SandakovMM: Unfortunately no. It causes `SyntaxError("cannot use absolute path on element")`

PR #70, opened 2024-09-04
Add an action to restart the plesk-firewalld service and disable the conflicting firewalld service

SandakovMM:


pleskdistup/actions/systemd.py
```diff
@@ -147,6 +147,37 @@ def estimate_revert_time(self) -> int:
...
+        util.logged_check_call(["/usr/bin/systemctl", "stop", "plesk-firewall.service"])
+        util.logged_check_call(["/usr/bin/systemctl", "disable", "plesk-firewall.service"])
+        return action.ActionResult()
+
+    def _post_action(self) -> action.ActionResult:
+        util.logged_check_call(["/usr/bin/systemctl", "enable", "plesk-firewall.service"])
+        # Don't do startup because the services will be started up after reboot at the end of the script anyway.
+        return action.ActionResult()
+
+    def _revert_action(self) -> action.ActionResult:
```
kpushkaryov, line 174: Maybe remember initial `firewalld` state and restore it on revert too?

kpushkaryov (changes requested):

pleskdistup/actions/systemd.py
SandakovMM: It's not obvious, because firewalld and plesk-firewall are not meant to operate at the same time. If both are enabled, something is likely wrong.
On the other hand centos2alma was not designed to fix such issues by reverting operations. Therefore, it might be better to return the server to its original state, even if that state is incorrect :thinking:

SandakovMM: Ah, wait a moment. Plesk-firewall already performs the same function. Therefore, this approach may not be the best. I will need to revise the change

kpushkaryov (approved):
Please rebase before merge.

PR #72, opened 2024-09-09
Removed lock file upon receiving interrupt signals

SandakovMM:


pleskdistup/main.py
```diff
@@ -317,6 +318,17 @@ def try_lock(lock_file: PathType) -> typing.Generator[bool, None, None]:
                 log.warn(f"Failed to remove lockfile {lock_file!r}: {ex}")
 
 
+def create_free_lock_exit_signal_handler(lock_file: PathType) -> typing.Callable:
+    def signal_handler(signum, frame):
+        log.info(f"Received signal {signum}, going to free lockfile {lock_file!r}...")
+        try:
+            os.unlink(lock_file)
```
kpushkaryov, line 325: I suspect it could cause double unlink with `try_lock()`, producing spurious warning in log, because `sys.exit()` will just raise `SystemExit` somewhere on the main thread, so `try_lock()` exit procedure will be executed too.

kpushkaryov (changes requested):

kpushkaryov (approved):
Please rebase before merge.

PR #75, opened 2024-09-23
FIX Check GRUB install_device's for inaccessible files (device)

umtk-webpros:
PAUX-6367: Distupgrade tool does not handle issue with GRUB misconfiguration

### DIFF
--- a/pleskdistup/actions/grub.py
+++ b/pleskdistup/actions/grub.py
@@ -1,0 +1,30 @@
+# Copyright 2023-2024. WebPros International GmbH. All rights reserved.
+import os.path
+
+from pleskdistup.common import action, util
+
+DEBCONF_CMD = "/usr/bin/debconf-show"
+
+class AssertGrubInstallDeviceExists(action.CheckAction):
+    def __init__(self) -> None:
+        self.name = "check GRUB installation device exists"
+        self.description = """Grub's install-device is not found: {}
+This could fail conversion - please fix GRUB configuration:
+    dpkg --configure grub-pc
+"""
+
+    def _do_check(self) -> bool:
+        if not os.path.exists(DEBCONF_CMD):
+            return True
+
+        param = "grub-pc"
+        debconf_out = util.logged_check_call([DEBCONF_CMD, param])
+        prefix = "{}/install_devices:".format(param)
+        for line in debconf_out.split("\n"):
+            idx = line.find(prefix)
+            if idx != -1:
+                devpath = line[idx+len(prefix):].strip()
+                if not os.path.exists(devpath):
+                    self.description = self.description.format(devpath)
+                    return False
+        return True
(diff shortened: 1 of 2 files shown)


pleskdistup/actions/grub.py
```diff
@@ -0,0 +1,27 @@
...
+
+from pleskdistup.common import action, dist, util
+
+DEBCONF_CMD = "/usr/bin/debconf-show"
+
+class AssertGrubInstallDeviceExists(action.CheckAction):
+    def __init__(self) -> None:
+        self.name = "check GRUB installation device exists"
+        self.description = """Grub's install-device is not found: {}
+This could fail conversion - please fix it.
```
SandakovMM, line 12: Can we propose a solution to fix this problem here, ideally using a bunch of CLI commands?

pleskdistup/actions/grub.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+
+    def _do_check(self) -> bool:
+        if not dist.get_distro().deb_based:
+            return True
+
+        debconf_out = util.logged_check_call([DEBCONF_CMD, "grub-pc"])
+        prefix = "grub-pc/install_devices:"
```
SandakovMM, line 20: NIT: We could turn this into a class member and make it configurable through the init function argument. However, it does not seem critical at the moment.

pleskdistup/actions/grub.py
```diff
@@ -0,0 +1,27 @@
...
+        for line in debconf_out.split("\n"):
+            if line.find(prefix) == 0:
+                devpath = line[len(prefix):].strip()
+                if not os.path.exists(devpath):
```
SandakovMM, line 24: Is it possible to have multiple `grub-pc/install_devices:` lines in the output?

ukablan-wpc: The virtual instance I encountered had only one device path. Although the plural noun suspects of multi device but I could not find a way to make it.

pleskdistup/actions/grub.py
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self) -> None:
+        self.name = "check GRUB installation device exists"
+        self.description = """Grub's install-device is not found: {}
+This could fail conversion - please fix GRUB configuration:
+    dpkg --configure grub-pc
+"""
+
+    def _do_check(self) -> bool:
+        if not os.path.exists(DEBCONF_CMD):
+            return True
```
ukablan-wpc, line 18: Better to check distro IMO

pleskdistup/actions/grub.py
ukablan-wpc: I added 
```
dpkg --configure grub-pc
```

pleskdistup/actions/grub.py
ukablan-wpc: Yes, it seems like (to stay) very debian-specific class anyway.

SandakovMM (approved):

PR #76, opened 2024-09-24
FIX Add ability to regexp-replace repo.list files

ukablan-wpc:
Fixes: PAUX-6369 Invalid APT sources produced by source switch procedure

### DIFF
--- a/pleskdistup/actions/distupgrade.py
+++ b/pleskdistup/actions/distupgrade.py
@@ -1,5 +1,6 @@
 # Copyright 2023-2024. WebPros International GmbH. All rights reserved.
 import os
+import re
 import subprocess
 import typing
 import urllib.request
@@ -143,6 +144,92 @@ class UpdateLegacyPhpRepositories(action.ActiveAction):
 
     def estimate_revert_time(self) -> int:
         return 20
+
+
+class ReplaceAptReposRegexp(action.ActiveAction):
+    from_regexp: str
+    to_regexp: str
+    sources_list_path: str
+    sources_list_d_path: str
+    _name: str
+
+    def __init__(
+        self,
+        from_regexp: str,
+        to_regexp: str,
+        sources_list_path: str = "/etc/apt/sources.list",
+        sources_list_d_path: str = "/etc/apt/sources.list.d/",
+        name: str = "set up APT repositories to change from {self.from_regexp!r} to {self.to_regexp!r}",
+    ) -> None:
+        self.from_regexp = from_regexp
+        self.to_regexp = to_regexp
+        self.sources_list_path = sources_list_path
+        self.sources_list_d_path = sources_list_d_path
+
+        self._name = name
+
+    @property
+    def name(self):
+        return self._name.format(self=self)
+
+    def _apply_replace_to_file(self, fpath: str, ptrn: re.Pattern, to_regexp: str) -> None:
+        changed = False
+        new_lines = []
+        with open(fpath) as f:
+            for line in f:
+                new_lines.append(ptrn.sub(to_regexp, line))
+                if new_lines[-1] != line:
+                    changed = True
+        if not changed:
+            return
+        files.backup_file(fpath)
+        with open(fpath, 'w') as f:
+            f.writelines(new_lines)
+
+    def _get_all_repo_list_files(self) -> typing.List[str]:
+        ret = [self.sources_list_path]
+        for root, _, filenames in os.walk(self.sources_list_d_path):
+            for f in filenames:
+                if f.endswith(".list"):
+                    ret.append(os.path.join(root, f))
+        return ret
+
+    def _rm_backups(self) -> None:
+        for f in self._get_all_repo_list_files():
+            files.remove_backup(f, log.debug)
+
+    def _change_by_regexp(self, from_regexp: str, to_regexp: str) -> None:
+        p = re.compile(from_regexp)
+        for f in self._get_all_repo_list_files():
+            self._apply_replace_to_file(f, p, to_regexp)
+
+    def _revert_all(self) -> None:
+        for f in self._get_all_repo_list_files():
+            files.restore_file_from_backup(f)
+
+    def _prepare_action(self) -> action.ActionResult:
+        self._change_by_regexp(self.from_regexp, self.to_regexp)
+        packages.update_package_list()
+        return action.ActionResult()
+
+    def _post_action(self) -> action.ActionResult:
+        self._rm_backups()
+        return action.ActionResult()
+
+    def _revert_action(self) -> action.ActionResult:
+        self._revert_all()
+        packages.update_package_list()
+        return action.ActionResult()
+
+    def estimate_prepare_time(self) -> int:
+        return 22
+
+    def estimate_revert_time(self) -> int:
+        return 22
+
+
+ReplaceAptReposRegexpDebian = ReplaceAptReposRegexp
+ReplaceAptReposRegexpUbuntu = ReplaceAptReposRegexp
 
 
 class SetupAptRepositories(action.ActiveAction):
--- a/pleskdistup/actions/distupgrade.py
+++ b/pleskdistup/actions/distupgrade.py
@@ -180,10 +180,12 @@ class CheckAptReposBackups(action.CheckAction):
         self._name = val
 
     def _do_check(self) -> bool:
-        log.debug(f"Checking leftover repo.list files")
-        archived_files = [f for f in
+        log.debug("Checking leftover repo.list files")
+        archived_files = [
+            f for f in
             CheckAptReposBackups.get_all_repo_list_files(self.sources_list_path, self.sources_list_d_path)
-                if files.backup_exists(f)]
+            if files.backup_exists(f)
+        ]
         if archived_files:
             self.description = self.description.format(",".join(archived_files))
             return False
@@ -235,8 +237,10 @@ class ReplaceAptReposRegexp(action.ActiveAction):
             f.writelines(new_lines)
 
     def _get_all_repo_list_files(self) -> typing.List[str]:
-        return CheckAptReposBackups.get_all_repo_list_files(self.sources_list_path,
-                    self.sources_list_d_path)
+        return CheckAptReposBackups.get_all_repo_list_files(
+            self.sources_list_path,
+            self.sources_list_d_path,
+        )
 
     def _rm_backups(self) -> None:
         for f in self._get_all_repo_list_files():
--- a/pleskdistup/actions/distupgrade.py
+++ b/pleskdistup/actions/distupgrade.py
@@ -239,7 +239,7 @@ class ReplaceAptReposRegexp(action.ActiveAction):
 
     def _rm_backups(self) -> None:
         for f in self._get_all_repo_list_files():
-            files.remove_backup(f, log.debug)
+            files.remove_backup(f, False, log.debug)
 
     def _change_by_regexp(self, from_regexp: str, to_regexp: str) -> None:
         p = re.compile(from_regexp)
--- a/pleskdistup/actions/distupgrade.py
+++ b/pleskdistup/actions/distupgrade.py
@@ -169,8 +169,12 @@ class ReplaceAptReposRegexp(action.ActiveAction):
         self._name = name
 
     @property
-    def name(self):
+    def name(self) -> str:
         return self._name.format(self=self)
+
+    @name.setter
+    def name(self, val: str) -> None:
+        self._name = val
 
     def _apply_replace_to_file(self, fpath: str, ptrn: re.Pattern, to_regexp: str) -> None:
         changed = False
--- a/pleskdistup/actions/distupgrade.py
+++ b/pleskdistup/actions/distupgrade.py
@@ -167,7 +167,10 @@ class CheckAptReposBackups(action.CheckAction):
         sources_list_d_path: str = "/etc/apt/sources.list.d/",
     ) -> None:
         self._name = "Check repo.list leftovers from previous conversion"
-        self.description = "These repo.list files are found with backups, please check/restore them, and delete the backups: {}"
+        self.description = """These repo.list backups are found:
+\t{}
+Please use `--revert` flag to restore and remove those backups.
+"""
         self.sources_list_path = sources_list_path
         self.sources_list_d_path = sources_list_d_path
 
--- a/pleskdistup/actions/distupgrade.py
+++ b/pleskdistup/actions/distupgrade.py
@@ -146,6 +146,49 @@ class UpdateLegacyPhpRepositories(action.ActiveAction):
         return 20
 
 
+class CheckAptReposBackups(action.CheckAction):
+    description: str
+    _name: str
+    sources_list_path: str
+    sources_list_d_path: str
+
+    @staticmethod
+    def get_all_repo_list_files(sources_list_path: str, sources_list_d_path: str) -> typing.List[str]:
+        ret = [sources_list_path]
+        for root, _, filenames in os.walk(sources_list_d_path):
+            for f in filenames:
+                if f.endswith(".list"):
+                    ret.append(os.path.join(root, f))
+        return ret
+
+    def __init__(self,
+        sources_list_path: str = "/etc/apt/sources.list",
+        sources_list_d_path: str = "/etc/apt/sources.list.d/",
+    ) -> None:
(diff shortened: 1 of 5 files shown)


pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +147,118 @@ def estimate_revert_time(self) -> int:
...
+        changed = False
+        new_lines = []
+        with open(fpath) as f:
+            for line in f:
+                new_lines.append(ptrn.sub(to_regexp, line))
+                if new_lines[-1] != line:
+                    changed = True
+        if not changed:
+            return
+        with open(fpath + self.backup_suffix, 'wb') as b:
```
SandakovMM, line 188: I suggest using [`files.backup_file`](https://github.com/plesk/dist-upgrader/blob/main/pleskdistup/common/src/files.py#L62-L64) to back up files.

ukablan-wpc:
> Overall, Could you please tell what the plan is for using regular expression substitution in the next step? Is it possible to avoid using regular expressions? =)

As the bug is related to repo.list lines i.e.
```
deb [arch=amd64,arm64,armhf] https://packages.microsoft.com/ubuntu/20.04/prod focal main
```
I am planning to utilize it to replace URLs with version strings inside (and possibly support smart replaces as opposed to `SetupAptRepositories`)

SandakovMM:
> I am planning to utilize it to replace URLs with version strings inside (and possibly support smart replaces as opposed to SetupAptRepositories)

I see.
I just want to mention that, IMO using regular expressions seems a bit excessive if we are only replacing "20" with "22". However, they are acceptable if we plan to add more complex change.

pleskdistup/actions/distupgrade.py
ukablan-wpc: I utilized it by changing a bit to match my logic. I didn't want to use some common `.bak` extensions so I produced my UUID.

ukablan-wpc: And I lost the ability to move my UUID suffixed temporary backups to some well known extension like ".distupg_save" at the end, just in case.

pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +147,97 @@ def estimate_revert_time(self) -> int:
...
+        sources_list_d_path: str = "/etc/apt/sources.list.d/",
+        name: str = "set up APT repositories to change from {self.from_regexp!r} to {self.to_regexp!r}",
+    ):
+        self.from_regexp = from_regexp
+        self.to_regexp = to_regexp
+        self.sources_list_path = sources_list_path
+        self.sources_list_d_path = sources_list_d_path
+
+        self._name = name
+        self.backup_suffix = "_" + str(uuid.uuid4())
```
SandakovMM, line 172: What is the purpose of configuring a backup suffix?

pleskdistup/common/src/files.py
```diff
@@ -59,25 +59,30 @@ def get_last_lines(filename: PathType, n: int) -> typing.List[str]:
         return f.readlines()[-n:]
 
 
-def backup_file(filename: str) -> None:
+def backup_file(filename: str, ext: str = ".bak") -> None:
     if os.path.exists(filename):
```
SandakovMM, line 63: I think we should check if the backup file already exists and skip copying it if it does. This prevents overwriting the backup file if it has already been created and the .list file has been changed

ukablan-wpc: But this will risk when user has a '.bak' long long times ago.

pleskdistup/common/src/files.py
```diff
@@ -59,25 +59,30 @@ def get_last_lines(filename: PathType, n: int) -> typing.List[str]:
...
 
-def remove_backup(filename: str) -> None:
-    if os.path.exists(filename + ".bak"):
-        os.remove(filename + ".bak")
+def remove_backup(filename: str, ext: str = ".bak", logf = None) -> None:
+    try:
+        if os.path.exists(filename + ext):
+            os.remove(filename + ext)
+    except Exception as ex:
+        if logf is None:
```
SandakovMM, line 83: I believe we could use `log.debug` here.

pleskdistup/actions/distupgrade.py
ukablan-wpc: Not to overwrite it with a well known suffix (like `.bak`)

SandakovMM, line 172: I believe uuid4 generates a unique UUID each time, so during the reverting/finishing stage, we will have different UUIDs and different suffixes

ukablan-wpc: Hmm, aren't they [using the same actions_map](https://github.com/plesk/dist-upgrader/blob/main/pleskdistup/main.py#L234)?

pleskdistup/common/src/files.py
SandakovMM: Let's then change it to a less common suffix like "conversion.back". I want to ensure we don't get overwhelmed by various suffixes for each case.

pleskdistup/actions/distupgrade.py
SandakovMM: Action objects will be recreated from scratch each time. They will perform the same tasks, but they will not be the same objects

ukablan-wpc: OK. This will make it impossible to keep some information in memory but, anyways, if this is a design choice.

pleskdistup/common/src/files.py
```diff
@@ -59,25 +59,31 @@ def get_last_lines(filename: PathType, n: int) -> typing.List[str]:
...
-def backup_file(filename: str) -> None:
+def backup_file(filename: str, ext: str = ".conversion.bak") -> None:
     if os.path.exists(filename):
-        shutil.copy(filename, filename + ".bak")
+        shutil.copy(filename, filename + ext)
 
 
-def backup_exists(filename: str) -> bool:
-    return os.path.exists(filename + ".bak")
+def backup_exists(filename: str, ext: str = ".conversion.bak") -> bool:
```
SandakovMM, line 67: There are several tests in [the motdtests.py](https://github.com/plesk/dist-upgrader/blob/main/pleskdistup/common/tests/motdtests.py#L9) should be fixed.
Additionally, it would be a good idea to add some tests related to backups. Up to you

pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +146,94 @@ def estimate_revert_time(self) -> int:
...
+    def _change_by_regexp(self, from_regexp: str, to_regexp: str) -> None:
+        p = re.compile(from_regexp)
+        self._apply_replace_on_file(self.sources_list_path, p, to_regexp)
+
+        for root, _, filenames in os.walk(self.sources_list_d_path):
+            for f in filenames:
+                if f.endswith(".list"):
+                    self._apply_replace_on_file(os.path.join(root, f), p, to_regexp)
+
+    def _revert_all(self):
```
kpushkaryov, line 205: `-> None`

pleskdistup/common/src/files.py
ukablan-wpc: I didn't want to break old behavior - throwing exception.

pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +146,94 @@ def estimate_revert_time(self) -> int:
         return 20
 
 
+class SetupAptReposRegexp(action.ActiveAction):
```
kpushkaryov, line 149: `SetupAptReposRegexp` doesn't exactly fit its purpose. Its more like `ReplaceAptReposRegexp`.

pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +146,94 @@ def estimate_revert_time(self) -> int:
...
+                    changed = True
+        if not changed:
+            return
+        files.backup_file(fpath)
+        with open(fpath, 'w') as f:
+            f.writelines(new_lines)
+
+    def _rm_backups(self):
+        files.remove_backup(self.sources_list_path, log.debug)
+        for root, _, filenames in os.walk(self.sources_list_d_path):
```
kpushkaryov, line 191: This nested loop is repeated 3 times, probably better to convert it to a generator or function.

pleskdistup/common/src/files.py
```diff
@@ -59,25 +60,31 @@ def get_last_lines(filename: PathType, n: int) -> typing.List[str]:
...
-    return os.path.exists(filename + ".bak")
+def backup_exists(filename: str, ext: str = DEFAULT_BACKUP_EXTENSION) -> bool:
+    return os.path.exists(filename + ext)
 
 
-def restore_file_from_backup(filename: str, remove_if_no_backup: bool = False) -> None:
-    if os.path.exists(filename + ".bak"):
-        shutil.move(filename + ".bak", filename)
+def restore_file_from_backup(filename: str, remove_if_no_backup: bool = False,
+                             ext: str = DEFAULT_BACKUP_EXTENSION) -> None:
```
kpushkaryov, line 73: Too much indent.

pleskdistup/common/src/files.py
```diff
@@ -59,25 +60,31 @@ def get_last_lines(filename: PathType, n: int) -> typing.List[str]:
...
+    if os.path.exists(filename + ext):
+        shutil.move(filename + ext, filename)
     elif remove_if_no_backup and os.path.exists(filename):
         os.remove(filename)
 
 
-def remove_backup(filename: str) -> None:
-    if os.path.exists(filename + ".bak"):
-        os.remove(filename + ".bak")
+def remove_backup(filename: str, logf = None, ext: str = DEFAULT_BACKUP_EXTENSION) -> None:
```
kpushkaryov, line 80: `logf` doesn't have type hint.

pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +146,94 @@ def estimate_revert_time(self) -> int:
...
+                new_lines.append(ptrn.sub(to_regexp, line))
+                if new_lines[-1] != line:
+                    changed = True
+        if not changed:
+            return
+        files.backup_file(fpath)
+        with open(fpath, 'w') as f:
+            f.writelines(new_lines)
+
+    def _rm_backups(self):
```
kpushkaryov, line 189: `-> None`

pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +146,94 @@ def estimate_revert_time(self) -> int:
...
+    _name: str
+
+    def __init__(
+        self,
+        from_regexp: str,
+        to_regexp: str,
+        sources_list_path: str = "/etc/apt/sources.list",
+        sources_list_d_path: str = "/etc/apt/sources.list.d/",
+        name: str = "set up APT repositories to change from {self.from_regexp!r} to {self.to_regexp!r}",
+    ):
```
kpushkaryov, line 163: `-> None`

kpushkaryov (changes requested):

pleskdistup/common/src/files.py
ukablan-wpc: It was automatic but looks good IMO, I will try 2 shifts

pleskdistup/common/src/files.py
ukablan-wpc: It would make it bigger, let's add test in a separate PR if it's OK?

pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +146,94 @@ def estimate_revert_time(self) -> int:
...
+        self.from_regexp = from_regexp
+        self.to_regexp = to_regexp
+        self.sources_list_path = sources_list_path
+        self.sources_list_d_path = sources_list_d_path
+
+        self._name = name
+
+    @property
+    def name(self):
```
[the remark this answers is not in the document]
ukablan-wpc: failed lint
```
pleskdistup/actions/distupgrade.py:172: error: Cannot override writeable attribute with read-only property  [override]
```
So I reverted this part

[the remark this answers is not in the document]
kpushkaryov: So, I advise to add a setter to the property to make it not read-only.

kpushkaryov (changes requested):
I advise to add a setter to the `name` property to make it not read-only. See the `InstallPackages` class for an example.

pleskdistup/common/src/files.py
SandakovMM: I suggest adding some logging and then raising an exception. Or if you want to avoid an exception in your specific case, we could add a boolean parameter to configure the function's behavior.

ukablan-wpc: I don't think there is a benefit ATM, both w/exception (as before) and logging (my current need) is working as we want (when one needs a behavior change later they'll do it).

SandakovMM: The issue I have is with the argument's name. In my opinion, it does not explain why it should be used or how it will affect the function's behavio

ukablan-wpc: Added explicit `raise_exception` param

pleskdistup/actions/distupgrade.py
```diff
@@ -145,6 +146,135 @@ def estimate_revert_time(self) -> int:
...
+                    changed = True
+        if not changed:
+            return
+        files.backup_file(fpath)
+        with open(fpath, 'w') as f:
+            f.writelines(new_lines)
+
+    def _get_all_repo_list_files(self) -> typing.List[str]:
+        return CheckAptReposBackups.get_all_repo_list_files(self.sources_list_path,
+                    self.sources_list_d_path)
```
kpushkaryov, line 238: Too much indent. I suggest opening a new line after '(' and use single indent afterwards, that is:
```python
return f(
    arg1,
    arg2,
)
```

pleskdistup/common/src/files.py
```diff
@@ -59,25 +60,35 @@ def get_last_lines(filename: PathType, n: int) -> typing.List[str]:
...
+        shutil.move(filename + ext, filename)
     elif remove_if_no_backup and os.path.exists(filename):
         os.remove(filename)
 
 
-def remove_backup(filename: str) -> None:
-    if os.path.exists(filename + ".bak"):
-        os.remove(filename + ".bak")
+def remove_backup(filename: str,
+        raise_exception: bool = True,
```
kpushkaryov, line 81: Too much indent. I suggest opening a new line after '(' and use single indent afterwards, that is:
```python
def f(
    arg1,
    arg2,
) -> None:
```

pleskdistup/common/src/files.py
```diff
@@ -59,25 +60,35 @@ def get_last_lines(filename: PathType, n: int) -> typing.List[str]:
...
+        logf : typing.Optional[typing.Callable] = None,
```
kpushkaryov, line 82: I advise more detailed specification instead of just `Callable`: `Callable[[str], Any]`.

kpushkaryov (changes requested):
Also please note failed flake8 checks.

ukablan-wpc:
> Also please note failed flake8 checks.

Only one, which seems an old code; should I touch it in this PR?

```
./pleskdistup/common/src/files.py:16:1: E302 expected 2 blank lines, found 1

File: pleskdistup/actions/distupgrade.py [hunk 17]
```diff
@@ -145,6 +146,140 @@ def estimate_revert_time(self) -> int:
...
+                    ret.append(os.path.join(root, f))
+        return ret
+
+    def __init__(
+        self,
+        sources_list_path: str = "/etc/apt/sources.list",
+        sources_list_d_path: str = "/etc/apt/sources.list.d/",
+    ) -> None:
+        self._name = "Check repo.list leftovers from previous conversion"
+        self.description = "These repo.list files are found with backups, please check/restore them, and delete the backups: {}"
```
SandakovMM, line 170: I think you could ask to use `--revert` here. It seems much simpler than manually searching for files.
In fact, it could be a great way to ensure the customer reverts after a failed conversion, thank you for the idea. :thinking:

SandakovMM: Everything else looks great, so let's change the description and merge this =)

ukablan-wpc: I wasn't sure if those archive files are purely safe so I asked the customer to check.
Asking about a command line switch in a library seems counter-intuitive to me but OK

SandakovMM (approved):

PR #77, opened 2024-09-25
Move Mailman systemd service operations to a separate action

SandakovMM:
We will now skip any interactions with the Mailman service if it is not active at the start of the conversion.
By default, mailman is not configured and the service will fail to start. Therefore, attempts to start it during the revert or finishing stage will also fail.

### DIFF
--- a/pleskdistup/actions/systemd.py
+++ b/pleskdistup/actions/systemd.py
@@ -83,6 +83,8 @@ class DisablePleskRelatedServicesDuringUpgrade(action.ActiveAction):
 
     def __init__(self) -> None:
         self.name = "disable plesk related services"
+        # Be cautious when adding the mailman service here. If mailman is not configured, the service will not start.
+        # The best way to handle mailman is to use the DisableServiceDuringUpgrade action.
         plesk_known_systemd_services = [
             "crond.service",
             "dovecot.service",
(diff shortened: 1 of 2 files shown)


pleskdistup/actions/systemd.py
```diff
@@ -176,6 +175,33 @@ def _revert_action(self) -> action.ActionResult:
...
+# We need to handle the mailman service separately because it is not configured and is stopped by default
+# If we try to start an unconfigured service, it will fail, causing issues during the revert and finish stages
+class HandleMailmanService(action.ActiveAction):
+    mailman_service: str
+
+    def __init__(self) -> None:
+        self.name = "handle mailman service"
+        self.mailman_service = "mailman.service"
+
+    def _is_required(self) -> bool:
```
ukablan-wpc, line 187: I suppose we are not calling prepare, post, revert if is_REquired is False, right?

SandakovMM: Yep, you right

pleskdistup/actions/systemd.py
```diff
@@ -176,6 +175,33 @@ def _revert_action(self) -> action.ActionResult:
         return action.ActionResult()
 
 
+# We need to handle the mailman service separately because it is not configured and is stopped by default
+# If we try to start an unconfigured service, it will fail, causing issues during the revert and finish stages
+class HandleMailmanService(action.ActiveAction):
+    mailman_service: str
+
```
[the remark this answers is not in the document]
SandakovMM: We don't need to declare "name" since it has already been declared in the base class [Action](https://github.com/plesk/dist-upgrader/blob/main/pleskdistup/common/src/action.py#L51-L53)

pleskdistup/actions/systemd.py
```diff
@@ -176,6 +175,33 @@ def _revert_action(self) -> action.ActionResult:
         return action.ActionResult()
 
 
+# We need to handle the mailman service separately because it is not configured and is stopped by default
+# If we try to start an unconfigured service, it will fail, causing issues during the revert and finish stages
+class HandleMailmanService(action.ActiveAction):
```
kpushkaryov, line 180: Seems that this action is not specific to Mailman. Can we just replace it with some generic `DisableServiceDuringUpgrade`?

kpushkaryov (changes requested):

SandakovMM:
Regarding the flake8 problem, it's strange. It seems to be checking out the master branch instead of the current branch (because my branch doesn't have grub.py at all). I will investigate it further when I have time

SandakovMM:
> Regarding the flake8 problem, it's strange. It seems to be checking out the master branch instead of the current branch (because my branch doesn't have grub.py at all). I will investigate it further when I have time

It appears that when the master branch adds a file, but the current branch does not, the file remains and triggers flake8 if it is bad. I'm not sure if this can be resolved within the scope of `actions/checkout`. Will create an issue

ukablan-wpc (approved):
LGTM

kpushkaryov:
> It appears that when the master branch adds a file, but the current branch does not, the file remains and triggers flake8 if it is bad. I'm not sure if this can be resolved within the scope of `actions/checkout`. Will create an issue

Workspace reuse?

PR #81, opened 2024-10-02
 Introduce a function to extract all includes from bind configuration file

SandakovMM:


pleskdistup/common/src/dns.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2023-2024. WebPros International GmbH. All rights reserved.
+import os
+from collections import deque
+
+
+def get_all_includes_from_bind_config(config_file: str) -> list:
```
vizovitin, line 6: `list[str]`

pleskdistup/common/src/dns.py
```diff
@@ -0,0 +1,26 @@
...
+    while queue:
+        current_file = queue.popleft()
+        if not os.path.exists(current_file):
+            continue
+
+        with open(current_file) as f:
+            for line in f:
+                line = line.strip()
+                if line.startswith("include"):
+                    include_file = line.split('"')[1].strip(";")
```
vizovitin, line 22: Why do you need `strip(";")` here?

pleskdistup/common/tests/dnstests.py
vizovitin, line 1: Propose to add some tests (or modify existing ones) with whitespaces in the line.

vizovitin (changes requested):

SandakovMM:
We need to support chroot because we store all named configurations in a chroot environment for named (/var/named/chroot/)

pleskdistup/common/src/dns.py
vizovitin: OK.
BTW, could you remind me the minimal Python version we're targeting?

pleskdistup/common/src/dns.py
```diff
@@ -1,14 +1,21 @@
 # Copyright 2023-2024. WebPros International GmbH. All rights reserved.
 import os
+import typing
 from collections import deque
 
 
-def get_all_includes_from_bind_config(config_file: str) -> list:
-    includes: list[str] = []
+def get_all_includes_from_bind_config(config_file: str, chroot: str = "") -> typing.List[str]:
```
vizovitin, line 7: I'd name this `chroot_dir` or `chroot_path`, but fine.

pleskdistup/common/src/dns.py
```diff
@@ -1,14 +1,21 @@
 # Copyright 2023-2024. WebPros International GmbH. All rights reserved.
 import os
+import typing
```
vizovitin, line 3: Probably easier to do `from typing import Deque, List`.

pleskdistup/common/src/dns.py
```diff
@@ -19,7 +26,13 @@ def get_all_includes_from_bind_config(config_file: str) -> list:
             for line in f:
                 line = line.strip()
                 if line.startswith("include"):
-                    include_file = line.split('"')[1].strip(";")
+                    include_file = line.split('"')[1]
+
+                    if include_file.startswith("/"):
+                        include_file = chroot + include_file
+                    else:
+                        include_file = os.path.join(os.path.dirname(current_file), include_file)
```
vizovitin, line 34: This is not a correct logic. For relative paths the file will be assumed to be in the directory previously nominated by a `directory` statement. In the absence of a `directory` option the directory from which bind was loaded is used.

Honestly, I see no reason to support this, as I don't think we use this.

pleskdistup/common/tests/dnstests.py
```diff
@@ -46,3 +110,50 @@ def test_nested_includes(self):
...
+    include "included.conf";
+    allow-recursion {
+        localnets;
+    };
+    directory "/var";
+    pid-file "/var/run/named/named.pid";
+};
+''')
+        self.assertEqual(dns.get_all_includes_from_bind_config("/test.conf", chroot="chroot"), ["chroot/included.conf"])
+
```
vizovitin, line 129: `shutil.rmtree("chroot")` here or (probably better) in `tearDown`?

pleskdistup/common/tests/dnstests.py
```diff
@@ -1,15 +1,24 @@
...
 
 import src.dns as dns
 
 
 class TestGetIncludeFromBindConfiguration(unittest.TestCase):
 
     def setUp(self):
-        with open("test.conf", "w") as actions_data_file:
-            actions_data_file.write('{ "actions": [] }')
+        with open("test.conf", "w") as test_bind_config:
```
vizovitin, line 12: Seems like this is used only by a single test.

pleskdistup/common/tests/dnstests.py
vizovitin: I didn't actually mean the spaces inside filenames, but spaces before / after / inside the directives.

vizovitin (changes requested):

pleskdistup/common/tests/dnstests.py
SandakovMM: I included tests that have spaces and tabs in the filename, before the directive, and within the directive itself

pleskdistup/common/src/dns.py
SandakovMM: You right; I assumed 'bind' works differently. Let's just block these types of paths and disregard them. It is unlikely that anyone has this kind of configuration. If they do, the conversion will simply stop due to the missing include file, which isn't a major issue.

pleskdistup/common/src/dns.py
SandakovMM: 3.6

pleskdistup/common/src/dns.py
```diff
@@ -1,21 +1,23 @@
...
 
+from . import log
 
-def get_all_includes_from_bind_config(config_file: str, chroot: str = "") -> typing.List[str]:
-    includes: typing.List[str] = []
+
+def get_all_includes_from_bind_config(config_file: str, chroot_dir: str = "") -> List[str]:
+    includes: List[str] = []
 
     if config_file.startswith("/"):
```
vizovitin, line 12: Nitpick: there's `os.path.isabs()` for that.

vizovitin (dismissed):
Please rebase before merge.

PR #82, opened 2024-10-02
Fix apt-get lock held by another process error by retry

ukablan-wpc:
Fixes PAUX-6388 Failed Upgrade - Could not get lock /var/cache/apt/archives/lock. It is held by process 37677 (python3)

### DIFF
--- a/pleskdistup/common/src/dpkg.py
+++ b/pleskdistup/common/src/dpkg.py
@@ -5,8 +5,11 @@ import os
 import re
 import subprocess
 import typing
+import time
 
-from . import files, util
+from . import files, log, util
+
+DPKG_TEMPFAIL_RETRY: typing.List[int] = [30, 60, 90, 120]
 
 APT_CHOOSE_OLD_FILES_OPTIONS = [
     "-o", "Dpkg::Options::=--force-confdef",
@@ -122,9 +125,59 @@ def safely_install_packages(
     install_packages(pkgs, repository, force_package_config)
 
 
+def _exec_retry_when_locked(
+    apt_get_cmd: typing.List[str],
+    tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None,
+    collect_stdout: bool = False,
+) -> str:
+    cant_get_lock = False
+    stdout = []
+
+    if tmpfail_retry_intervals is None:
+        tmpfail_retry_intervals = DPKG_TEMPFAIL_RETRY
+
+    def process_stdout(line: str) -> None:
+        if collect_stdout:
+            nonlocal stdout
+            stdout.append(line)
+        log.info("stdout: {}".format(line.rstrip('\n')), to_stream=False)
+
+    def process_stderr(line: str) -> None:
+        log.info("stderr: {}".format(line.rstrip('\n')))
+        nonlocal cant_get_lock
+        if cant_get_lock:
+            return
+        if "E: Could not get lock" in line:
+            cant_get_lock = True
+
+    i = 0
+    while True:
+        cant_get_lock = False
+        stdout.clear()
+        log.info(f"Executing: {' '.join(apt_get_cmd)}")
+        exit_code = util.exec_get_output_streamed(
+            apt_get_cmd, process_stdout, process_stderr,
+            env={
+                "PATH": os.environ["PATH"],
+                "DEBIAN_FRONTEND": "noninteractive",
+                "LC_ALL": "C",
+                "LANG": "C",
+            },
+        )
+        if exit_code == 0:
+            break
+        if i >= len(tmpfail_retry_intervals) or not cant_get_lock:
+            raise subprocess.CalledProcessError(returncode=exit_code, cmd=apt_get_cmd)
+        log.info(f"{apt_get_cmd[0]} failed because lock is already held, will retry in {tmpfail_retry_intervals[i]} seconds..")
+        time.sleep(tmpfail_retry_intervals[i])
+        i += 1
+    return "".join(stdout)
+
+
 def remove_packages(
     pkgs: typing.List[str],
     simulate: bool = False,
+    tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None,
 ) -> typing.Optional[typing.Dict[str, typing.List[PackageEntry]]]:
     if len(pkgs) == 0:
         return None
@@ -133,7 +186,7 @@ def remove_packages(
     if simulate:
         cmd.append("--simulate")
     cmd += pkgs
-    cmd_out = util.logged_check_call(cmd)
+    cmd_out = _exec_retry_when_locked(cmd, tmpfail_retry_intervals, collect_stdout=True)
     if simulate:
         return _parse_apt_get_simulation(cmd_out)
     return None
@@ -142,6 +195,7 @@ def remove_packages(
 def safely_remove_packages(
     pkgs: typing.List[str],
     protected_pkgs: typing.Optional[typing.Iterable[str]] = None,
+    tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None,
 ) -> None:
     sim_res = remove_packages(pkgs, simulate=True)
     if sim_res is not None and protected_pkgs is not None:
@@ -149,28 +203,32 @@ def safely_remove_packages(
         violations = _find_protection_violations(sim_res, protected_set)
         if violations:
             raise PackageProtectionError(protected_packages=violations)
-    remove_packages(pkgs)
+    remove_packages(pkgs, False, tmpfail_retry_intervals)
 
 
 def find_related_repofiles(repository_file: str) -> typing.List[str]:
     return files.find_files_case_insensitive("/etc/apt/sources.list.d", repository_file)
 
 
-def update_package_list() -> None:
-    util.logged_check_call(["/usr/bin/apt-get", "update", "-y"])
+def update_package_list(tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None) -> None:
+    cmd = ["/usr/bin/apt-get", "update", "-y"]
+    _exec_retry_when_locked(cmd, tmpfail_retry_intervals)
 
 
-def upgrade_packages(pkgs: typing.Optional[typing.List[str]] = None) -> None:
+def upgrade_packages(
+    pkgs: typing.Optional[typing.List[str]] = None,
+    tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None,
+) -> None:
     if pkgs is None:
         pkgs = []
 
     cmd = ["/usr/bin/apt-get", "upgrade", "-y"] + APT_CHOOSE_OLD_FILES_OPTIONS + pkgs
-    util.logged_check_call(cmd, env={"PATH": os.environ["PATH"], "DEBIAN_FRONTEND": "noninteractive"})
+    _exec_retry_when_locked(cmd, tmpfail_retry_intervals)
 
 
-def autoremove_outdated_packages() -> None:
-    util.logged_check_call(["/usr/bin/apt-get", "autoremove", "-y"],
-                           env={"PATH": os.environ["PATH"], "DEBIAN_FRONTEND": "noninteractive"})
+def autoremove_outdated_packages(tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None) -> None:
+    cmd = ["/usr/bin/apt-get", "autoremove", "-y"]
+    _exec_retry_when_locked(cmd, tmpfail_retry_intervals)
 
 
 def depconfig_parameter_set(parameter: str, value: str) -> None:
@@ -184,13 +242,14 @@ def depconfig_parameter_get(parameter: str) -> str:
     return process.stdout.split(" ")[1].strip()
 
 
-def restore_installation() -> None:
-    util.logged_check_call(["/usr/bin/apt-get", "-f", "install", "-y"])
+def restore_installation(tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None) -> None:
+    cmd = ["/usr/bin/apt-get", "-f", "install", "-y"]
+    _exec_retry_when_locked(cmd, tmpfail_retry_intervals)
 
 
-def do_distupgrade() -> None:
-    util.logged_check_call(["apt-get", "dist-upgrade", "-y"] + APT_CHOOSE_OLD_FILES_OPTIONS,
-                           env={"PATH": os.environ["PATH"], "DEBIAN_FRONTEND": "noninteractive"})
+def do_distupgrade(tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None) -> None:
+    cmd = ["apt-get", "dist-upgrade", "-y"] + APT_CHOOSE_OLD_FILES_OPTIONS
+    _exec_retry_when_locked(cmd, tmpfail_retry_intervals)
 
 
 def get_installed_packages_list(regex: str) -> typing.List[typing.Tuple[str, str]]:
--- a/pleskdistup/common/src/util.py
+++ b/pleskdistup/common/src/util.py
@@ -6,34 +6,74 @@ import typing
 from . import log
 
 
-# Returns standard output
-def logged_check_call(cmd: typing.Union[typing.Sequence[str], str], **kwargs) -> str:
+def log_outputs_check_call(
+    cmd: typing.Union[typing.Sequence[str], str],
+    collect_return_stdout: bool = False,
+    **kwargs,
+) -> str:
+    '''
+    Runs cmd and raises on nonzero exit code. Returns stdout when collect_return_stdout
+    '''
     log.info(f"Running: {cmd!r}. Output:")
-
-    # I beleive we should be able pass argument to the subprocess function
-    # from the caller. So we have to inject stdout/stderr/universal_newlines
-    kwargs["stdout"] = subprocess.PIPE
-    kwargs["stderr"] = subprocess.STDOUT
-    kwargs["universal_newlines"] = True
-
     stdout = []
-    process = subprocess.Popen(cmd, **kwargs)
-    while None is process.poll():
-        if not process.stdout:
-            log.err(f"Can't get process output from {cmd!r}")
-            raise RuntimeError(f"Can't get process output from {cmd!r}")
-        line = process.stdout.readline()
-        if line:
-            stdout.append(line)
-            if line.strip():
-                log.info(line.strip(), to_stream=False)
 
-    if process.returncode != 0:
-        log.err(f"Command {cmd!r} failed with return code {process.returncode}")
-        raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd, output="\n".join(stdout))
+    def proc_stdout(line: str) -> None:
+        log.info("stdout: {}".format(line.rstrip('\n')), to_stream=False)
+        if collect_return_stdout:
+            stdout.append(line)
+
+    def proc_stderr(line: str) -> None:
+        log.info("stderr: {}".format(line.rstrip('\n')))
+
+    exit_code = exec_get_output_streamed(cmd, proc_stdout, proc_stderr, **kwargs)
+    if exit_code != 0:
+        log.err(f"Command {cmd!r} failed with return code {exit_code}")
+        raise subprocess.CalledProcessError(returncode=exit_code, cmd=cmd)
 
     log.info(f"Command {cmd!r} finished successfully")
-    return "\n".join(stdout)
+    return "".join(stdout)
+
+
+# Returns standard output
+def logged_check_call(cmd: typing.Union[typing.Sequence[str], str], **kwargs) -> str:
+    return log_outputs_check_call(cmd, collect_return_stdout=True, **kwargs)
+
+
+def exec_get_output_streamed(
+    cmd: typing.Union[typing.Sequence[str], str],
+    process_stdout_line: typing.Optional[typing.Callable[[str], None]],
+    process_stderr_line: typing.Optional[typing.Callable[[str], None]],
+    **kwargs,
+) -> int:
+    '''
+    Allows to get stdout/stderr by streaming line by line, by calling callbacks
+    and returns process exit code
+    '''
+    kwargs["stdout"] = (subprocess.DEVNULL if process_stdout_line is None
+                        else subprocess.PIPE)
+    kwargs["stderr"] = (subprocess.DEVNULL if process_stderr_line is None
+                        else subprocess.PIPE)
+    kwargs["universal_newlines"] = True
+
+    process = subprocess.Popen(cmd, **kwargs)
+    if process_stdout_line is None and process_stderr_line is None:
+        process.communicate()
+        return process.returncode
+
+    while process.poll() is None:
+        if process_stdout_line is not None:
+            if not process.stdout:
+                raise RuntimeError(f"Cannot get process stdout of command {cmd!r}")
+            line = process.stdout.readline()
+            if line:
+                process_stdout_line(line)
+        if process_stderr_line is not None:
+            if not process.stderr:
+                raise RuntimeError(f"Cannot get process stderr of command {cmd!r}")
+            line = process.stderr.readline()
+            if line:
+                process_stderr_line(line)
+    return process.returncode
 
 
 def merge_dicts_of_lists(


pleskdistup/common/src/util.py
vizovitin, line 1: Did you check that this doesn't break all existing uses of this function?

pleskdistup/common/src/util.py
```diff
@@ -10,13 +10,14 @@
...
     kwargs["stdout"] = subprocess.PIPE
-    kwargs["stderr"] = subprocess.STDOUT
+    kwargs["stderr"] = subprocess.PIPE
     kwargs["universal_newlines"] = True
 
     stdout = []
+    stderr = []
     process = subprocess.Popen(cmd, **kwargs)
     while None is process.poll():
         if not process.stdout:
```
vizovitin, line 23: Why is this logic asymmetric? Will this work correctly if the process only outputs to stderr (and a lot)?

pleskdistup/common/src/util.py
```diff
@@ -10,13 +10,14 @@
...
     # from the caller. So we have to inject stdout/stderr/universal_newlines
     kwargs["stdout"] = subprocess.PIPE
-    kwargs["stderr"] = subprocess.STDOUT
+    kwargs["stderr"] = subprocess.PIPE
     kwargs["universal_newlines"] = True
 
     stdout = []
+    stderr = []
     process = subprocess.Popen(cmd, **kwargs)
     while None is process.poll():
```
vizovitin, line 22: Why not simply use something like:
```py
    out, err = process.communicate()
```

pleskdistup/common/src/dpkg.py
```diff
@@ -188,9 +190,23 @@ def restore_installation() -> None:
...
+    while True:
+        try:
+            util.logged_check_call(
+                ["apt-get", "dist-upgrade", "-y"] + APT_CHOOSE_OLD_FILES_OPTIONS,
+                env={"PATH": os.environ["PATH"], "DEBIAN_FRONTEND": "noninteractive"},
+            )
+        except subprocess.CalledProcessError as e:
+            if i >= len(locked_sleep_intervals):
+                raise
+            if "Could not get lock" in e.stderr:
```
vizovitin, line 204: `"E: Could not get lock"`

pleskdistup/common/src/dpkg.py
```diff
@@ -188,9 +190,23 @@ def restore_installation() -> None:
...
+                log.info(f"apt-get exited with locking problem, will retry in {locked_sleep_intervals[i]} seconds..")
```
vizovitin, line 205: It's not a locking problem, it's just being locked by another process.

vizovitin (changes requested):

pleskdistup/common/src/dpkg.py
```diff
@@ -188,9 +190,36 @@ def restore_installation() -> None:
...
+    def process_stderr(line: str) -> None:
+        log.err(line)
+        nonlocal cant_get_lock
+        if cant_get_lock:
+            return
+        if "E: Could not get lock" in line:
+            cant_get_lock = True
+
+    cmd = ["apt-get", "dist-upgrade", "-y"] + APT_CHOOSE_OLD_FILES_OPTIONS
+    while True:
```
SandakovMM, line 209: Can we use `for interval in locked_sleep_intervals` instead? This way, we can at least get rid of `i` and check for it

pleskdistup/common/src/util.py
```diff
@@ -10,32 +10,65 @@
...
+                raise RuntimeError(f"Cannot get process stdout of command {cmd!r}")
+            line = process.stdout.readline()
+            process_stdout_line(line)
+        if process_stderr_line is not None:
+            if not process.stderr:
+                raise RuntimeError(f"Cannot get process stderr of command {cmd!r}")
+            line = process.stderr.readline()
+            process_stderr_line(line)
+    return process.returncode
+
```
ukablan-wpc, line 70: Tried the cases:
```
    print(exec_get_output_streamed(["ls", "/aa"], None, None))
    print(exec_get_output_streamed("ls", None, None))

File: pleskdistup/common/src/util.py (hunk 2 above)
Comment (SandakovMM) [reply pleskdistup/common/src/util.py]: IIRC when we use `process.communicate`, we will be stuck until the subprocess is finished, right?

File: pleskdistup/common/src/dpkg.py (hunk 5 above)
Comment (ukablan-wpc) [reply pleskdistup/common/src/dpkg.py:209]: Yes it's my first thought as well but `locked_sleep_intervals` could be empty which means no retry at all - so I needed `do ... while()` in python

File: pleskdistup/common/src/util.py [hunk 7]
```diff
@@ -10,32 +10,65 @@
...
+    else:
+        kwargs["stdout"] = subprocess.PIPE
+    if process_stderr_line is None:
+        kwargs["stderr"] = subprocess.DEVNULL
+    else:
+        kwargs["stderr"] = subprocess.PIPE
+    kwargs["universal_newlines"] = True
+
+    process = subprocess.Popen(cmd, **kwargs)
+    while None is process.poll():
```
ukablan-wpc, line 58: Note: High CPU usage possibility when both are `None`?

pleskdistup/common/src/util.py
ukablan-wpc: AFAIS we're already stuck until the process terminates.
The difference would be, if we care, when the outputs are huge. I changed the logic to line process both outputs.

SandakovMM: We need to read the output and write it into the log file asynchronously. This is important for long-running processes like 'leapp upgrade' so we can log output while the process is still running. Some customers have used this before to confirm that the process is still active, even after the progress bar shows 100% (due to previous time estimate errors). I would like to maintain this functionality.

pleskdistup/common/src/dpkg.py
SandakovMM: Perhaps we should create our own simple decorator for this, as the default one does not support any delay.

pleskdistup/common/src/util.py
ukablan-wpc: Both `process.communicate()` and our current approach doesn't let the upgrader program proceed before child process it terminated. Was it already not what you expect?

pleskdistup/common/src/util.py
ukablan-wpc:
```
2024-10-03 12:33:07,651 - INFO - Do: upgrade packages
2024-10-03 12:33:07,652 - INFO - Executing: /usr/bin/apt-get upgrade -y -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold
2024-10-03 12:33:07,685 - INFO - 
2024-10-03 12:33:07,685 - ERROR - E: Could not get lock /var/lib/dpkg/lock. It is held by process 3638 (lcck)

2024-10-03 12:33:07,686 - INFO - dist-upgrade failed because lock is already held, will retry in 30 seconds..
2024-10-03 12:33:37,716 - INFO - Executing: /usr/bin/apt-get upgrade -y -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold
2024-10-03 12:33:39,976 - INFO - Reading package lists...

2024-10-03 12:33:43,843 - ERROR - 
2024-10-03 12:33:43,844 - INFO - Executing: /usr/bin/apt-get autoremove -y
2024-10-03 12:33:43,878 - INFO - Reading package lists...

2024-10-03 12:33:44,670 - ERROR - 
2024-10-03 12:33:44,670 - DEBUG - Action 'upgrade packages' completed in 37 s with result: ActionResult(state=<ActionState.SUCCESS: 'success'>, info=None, reboot_requested=None, next_phase=None)
2024-10-03 12:33:44,671 - INFO - Success: upgrade packages

Comment (ukablan-wpc) [reply pleskdistup/common/src/util.py:1]: Fixed the empty error line case by adding empty check before logging it.

File: pleskdistup/common/src/util.py (hunk 1 above)
Comment (ukablan-wpc) [reply pleskdistup/common/src/util.py]: Changed the logic to line by line consumption, please recheck

File: pleskdistup/common/src/dpkg.py (hunk 5 above)
Comment (ukablan-wpc) [reply pleskdistup/common/src/dpkg.py]: Yes maybe later. Our need is a bit different though: Feed the ("command") decorator with (possibly many) regexp/strings that the program might emit and is suitable for temporary failure.

File: pleskdistup/common/src/util.py
Comment (ukablan-wpc) [reply pleskdistup/common/src/util.py:1]:
```
2024-10-03 13:13:07,124 - INFO - Do: upgrade packages
2024-10-03 13:13:07,125 - INFO - Executing: /usr/bin/apt-get upgrade -y -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold
2024-10-03 13:13:07,170 - ERROR - E: Could not get lock /var/lib/dpkg/lock. It is held by process 1947 (lcck)

2024-10-03 13:13:07,170 - INFO - dist-upgrade failed because lock is already held, will retry in 30 seconds..
2024-10-03 13:13:37,201 - INFO - Executing: /usr/bin/apt-get upgrade -y -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold
2024-10-03 13:13:37,458 - INFO - Reading package lists...

2024-10-03 13:13:42,875 - INFO - Executing: /usr/bin/apt-get autoremove -y
2024-10-03 13:13:42,914 - INFO - Reading package lists...

2024-10-03 13:13:44,011 - DEBUG - Action 'upgrade packages' completed in 37 s with result: ActionResult(state=<ActionState.SUCCESS: 'success'>, info=None, reboot_requested=None, next_phase=None)
2024-10-03 13:13:44,012 - INFO - Success: upgrade packages

pleskdistup/common/src/util.py
ukablan-wpc: Yes it was, fixed.

pleskdistup/common/src/util.py
ukablan-wpc: Just another log:
```
2024-10-04 10:19:04,496 - INFO - Do: upgrade packages
2024-10-04 10:19:04,496 - INFO - Executing: /usr/bin/apt-get upgrade -y -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold
2024-10-04 10:19:04,515 - ERROR - E: Could not get lock /var/lib/dpkg/lock. It is held by process 2637 (lcck)
2024-10-04 10:19:04,516 - INFO - dist-upgrade failed because lock is already held, will retry in 30 seconds..
2024-10-04 10:19:34,546 - INFO - Executing: /usr/bin/apt-get upgrade -y -o Dpkg::Options::=--force-confdef -o Dpkg::Options::=--force-confold
2024-10-04 10:19:34,584 - INFO - Reading package lists...
2024-10-04 10:19:35,624 - INFO - Executing: /usr/bin/apt-get autoremove -y
2024-10-04 10:19:35,664 - INFO - Reading package lists...
2024-10-04 10:19:36,797 - DEBUG - Action 'upgrade packages' completed in 32 s with result: ActionResult(state=<ActionState.SUCCESS: 'success'>, info=None, reboot_requested=None, next_phase=None)
2024-10-04 10:19:36,799 - INFO - Success: upgrade packages
2024-10-04 10:19:36,799 - DEBUG - Invoking action 'update Plesk'
2024-10-04 10:19:36,799 - INFO - Do: update Plesk
2024-10-04 10:19:36,799 - INFO - Running: ['/usr/sbin/plesk', 'installer', 'update']. Output:
...

File: pleskdistup/common/src/dpkg.py [hunk 8]
```diff
@@ -122,9 +126,54 @@ def safely_install_packages(
...
+        tmpfail_retry_intervals = DPKG_TEMPFAIL_RETRY
+
+    def process_stdout(line: str) -> None:
+        if collect_stdout:
+            nonlocal stdout
+            stdout.append(line)
+        log.info(line)
+
+    def process_stderr(line: str) -> None:
+        log.err(line)
```
vizovitin, line 147: Questionable that this is an error in terms of log.

pleskdistup/common/src/dpkg.py
```diff
@@ -122,9 +126,54 @@ def safely_install_packages(
...
+        nonlocal cant_get_lock
+        if cant_get_lock:
+            return
+        if "E: Could not get lock" in line:
```
vizovitin, line 151: You probably should ensure a specific locale is set via environment.

pleskdistup/common/src/dpkg.py
```diff
@@ -122,9 +126,54 @@ def safely_install_packages(
...
+    while True:
+        log.info(f"Executing: {' '.join(apt_get_cmd)}")
+        exit_code = util.exec_get_output_streamed(
+            apt_get_cmd, process_stdout, process_stderr,
+            env={"PATH": os.environ["PATH"], "DEBIAN_FRONTEND": "noninteractive"},
+        )
+        if exit_code == 0:
+            break
+        if i >= len(tmpfail_retry_intervals) or not cant_get_lock:
+            raise subprocess.CalledProcessError(returncode=exit_code, cmd=apt_get_cmd)
```
vizovitin, line 164: What about stdout and stderr fields?

pleskdistup/common/src/dpkg.py
```diff
@@ -122,9 +126,54 @@ def safely_install_packages(
...
+        log.info(f"dist-upgrade failed because lock is already held, will retry in {tmpfail_retry_intervals[i]} seconds..")
```
vizovitin, line 165: Why "dist-upgrade"? Did it really fail if we're going to retry?

pleskdistup/common/src/dpkg.py
```diff
@@ -122,9 +126,54 @@ def safely_install_packages(
...
+        time.sleep(tmpfail_retry_intervals[i])
+        i += 1
+        cant_get_lock = False
+        stdout.clear()
```
vizovitin, line 169: Is stderr lost intentionally?

pleskdistup/common/src/dpkg.py
```diff
@@ -122,9 +126,54 @@ def safely_install_packages(
     install_packages(pkgs, repository, force_package_config)
 
 
+def apt_get_retry_temp_fails(
+    apt_get_cmd: typing.List[str],
+    tmpfail_retry_intervals: typing.Optional[typing.List[int]],
```
SandakovMM, line 131: I believe we want default value `=None` for the argument, don't we?

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,62 @@
...
-    if process.returncode != 0:
-        log.err(f"Command {cmd!r} failed with return code {process.returncode}")
-        raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd, output="\n".join(stdout))
+    if process_stdout_line is None and process_stderr_line is None:
+        process.communicate()
+        return process.returncode
 
-    log.info(f"Command {cmd!r} finished successfully")
-    return "\n".join(stdout)
+    while None is process.poll():
```
vizovitin, line 54: Why is `None` on the left?

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,62 @@
...
+        if process_stdout_line is not None:
+            if not process.stdout:
+                raise RuntimeError(f"Cannot get process stdout of command {cmd!r}")
+            line = process.stdout.readline()
+            if line:
+                process_stdout_line(line.rstrip())
```
vizovitin, line 60: Why do modify the line (remove all whitespace at the end)?

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,62 @@
 # Returns standard output
 def logged_check_call(cmd: typing.Union[typing.Sequence[str], str], **kwargs) -> str:
     log.info(f"Running: {cmd!r}. Output:")
+    stdout = []
+
+    def proc_stdout(line: str) -> None:
+        stdout.append(line)
```
SandakovMM, line 15: We should also write it into the `log.info` as it was before the change

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,62 @@
...
-            stdout.append(line)
-            if line.strip():
-                log.info(line.strip(), to_stream=False)
-
-    if process.returncode != 0:
-        log.err(f"Command {cmd!r} failed with return code {process.returncode}")
-        raise subprocess.CalledProcessError(returncode=process.returncode, cmd=cmd, output="\n".join(stdout))
+    if process_stdout_line is None and process_stderr_line is None:
+        process.communicate()
+        return process.returncode
```
vizovitin, line 52: Now this doesn't raise on non-zero result. This is a significant behavior change.

vizovitin (changes requested):

pleskdistup/common/src/util.py
ukablan-wpc: Behavior change could only be about previous function `logged_check_call()` which already [raises when exit_code is nonzero](https://github.com/plesk/dist-upgrader/pull/82/files#diff-7da365a7435ef4d10f5c4359e026a138b569ad1448f15af07a6166e72c07d9a9R21)

pleskdistup/common/src/util.py
ukablan-wpc: I don't have a strong opinion about it so I kept as it was.

pleskdistup/common/src/util.py
ukablan-wpc: Yes it was a tricky decision that I had to take. For the "purity" of the API, we should not remove any byte from the stream but this means that removing the end newline at the caller side. I chose to simplify caller function especially if it log.xxx(line).
If you think it should be the opposite and caller should know that it'll receive the line as is (with newline character) then it'll be OK for me although it'll add one more strip() line in callback fun.

pleskdistup/common/src/dpkg.py
ukablan-wpc: Thanks, it was wrong name

pleskdistup/common/src/dpkg.py
ukablan-wpc: I don't think we need them in the exception as they are already logged

pleskdistup/common/src/dpkg.py
ukablan-wpc: If the apt process is failing and printing to stderr, for me it's naturally ERR log. However we print in info level that it's OK and will be retried

pleskdistup/common/src/dpkg.py
```diff
@@ -122,9 +126,54 @@ def safely_install_packages(
...
+    collect_stdout: bool = False
+) -> str:
+    cant_get_lock = False
+    stdout = []
+
+    if tmpfail_retry_intervals is None:
```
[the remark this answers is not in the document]
ukablan-wpc: This is only used internally but I'll add it.

pleskdistup/common/src/util.py
ukablan-wpc: I just remembered C/C++ programmers decide to put the literal on the left side of the if expression just not to fall in the trap of single `=`. I personally read the other way around better but I am used to both variants ;)

pleskdistup/common/src/util.py
ukablan-wpc: Just to give more detail:
`logged_check_call` utilizes `exec_get_output_streamed` and its behavior should be intact. However we use `exec_get_output_streamed` to retry transient failures in dpkg* commands.

pleskdistup/common/src/dpkg.py
vizovitin: apt-get is not the only process to be launched using these functions.

pleskdistup/common/src/dpkg.py
vizovitin: I think it's better to set `LANG` or both.

pleskdistup/common/src/dpkg.py
vizovitin: Why not? I think a typical user would expect as much.

pleskdistup/common/src/util.py
vizovitin: This is a parasitic habit of PHP programmers. There's certainly no need for it in Python.

pleskdistup/common/src/util.py
vizovitin: Why not `.rstrip('\n')` instead?

pleskdistup/common/src/dpkg.py
ukablan-wpc: Previously it only included stdout in the exception but stderr was merely logged.
If we need both of them, it could be an issue if lots of output needs to be captured. Note that I changed it to stream processing the stdout/err because of this possibility.

pleskdistup/common/src/dpkg.py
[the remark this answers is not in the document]
vizovitin: If this function is intended to be used only internally, it should start with `_`.

pleskdistup/common/src/dpkg.py
vizovitin: You still accumulate the entire stdout, so this point is moot.
Anyway, I'm just pointing out API inconsistency. You could probably ignore that for now.

pleskdistup/common/src/util.py
ukablan-wpc: Done.

pleskdistup/common/src/util.py
ukablan-wpc: Swapped

pleskdistup/common/src/dpkg.py
ukablan-wpc: Added

pleskdistup/common/src/dpkg.py
ukablan-wpc: Changed func name if it's enough?

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,63 @@
...
+        log.err(f"Command {cmd!r} failed with return code {exit_code}")
+        raise subprocess.CalledProcessError(returncode=exit_code, cmd=cmd, output="\n".join(stdout))
+
+    log.info(f"Command {cmd!r} finished successfully")
+    return "\n".join(stdout)
 
-    # I beleive we should be able pass argument to the subprocess function
+
+# Allows to get stdout/stderr by streaming line by line, calling callbacks
+# and returns process exit code
```
vizovitin, line 31: There are function documentation strings for that.

pleskdistup/common/src/dpkg.py
ukablan-wpc: Correction: it's [logged as info()](https://github.com/plesk/dist-upgrader/pull/82/files#diff-4715ec6681042fd8dd5b835f6271b7caade7ccd3e11351aba43b84f4ee706306R144)

pleskdistup/common/src/dpkg.py
```diff
@@ -122,9 +126,58 @@ def safely_install_packages(
...
+        )
+        if exit_code == 0:
+            break
+        if i >= len(tmpfail_retry_intervals) or not cant_get_lock:
+            raise subprocess.CalledProcessError(returncode=exit_code, cmd=apt_get_cmd)
+        log.info(f"{apt_get_cmd[0]} failed because lock is already held, will retry in {tmpfail_retry_intervals[i]} seconds..")
+        time.sleep(tmpfail_retry_intervals[i])
+        i += 1
+        cant_get_lock = False
+        stdout.clear()
```
vizovitin, line 174: Nitpick: I think it would make more sense to initialize/reset these 2 at the start of the loop (before `util.exec_get_output_streamed()`).

pleskdistup/common/src/dpkg.py
```diff
@@ -184,13 +242,14 @@ def depconfig_parameter_get(parameter: str) -> str:
...
-    util.logged_check_call(["/usr/bin/apt-get", "-f", "install", "-y"])
+def restore_installation(tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None) -> None:
+    cmd = ["/usr/bin/apt-get", "-f", "install", "-y"]
+    apt_get_retry_temp_fails(cmd, tmpfail_retry_intervals)
 
 
-def do_distupgrade() -> None:
-    util.logged_check_call(["apt-get", "dist-upgrade", "-y"] + APT_CHOOSE_OLD_FILES_OPTIONS,
-                           env={"PATH": os.environ["PATH"], "DEBIAN_FRONTEND": "noninteractive"})
+def do_distupgrade(tmpfail_retry_intervals: typing.Optional[typing.List[int]] = None) -> None:
```
vizovitin, line 251: Do we really need to expose `tmpfail_retry_intervals` everywhere? What for?

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,63 @@
 # Returns standard output
 def logged_check_call(cmd: typing.Union[typing.Sequence[str], str], **kwargs) -> str:
     log.info(f"Running: {cmd!r}. Output:")
```
vizovitin, line 11: So basically if the logging level is above INFO, then the log will just contain some stderr strings w/o actually referencing the command that was executed.

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,63 @@
...
+    def proc_stderr(line: str) -> None:
+        log.err(line)
+
+    exit_code = exec_get_output_streamed(cmd, proc_stdout, proc_stderr, **kwargs)
+    if exit_code != 0:
+        log.err(f"Command {cmd!r} failed with return code {exit_code}")
+        raise subprocess.CalledProcessError(returncode=exit_code, cmd=cmd, output="\n".join(stdout))
+
+    log.info(f"Command {cmd!r} finished successfully")
+    return "\n".join(stdout)
```
vizovitin, line 27: Nitpick: note that in general that would not be equivalent to the real process stdout (due to some potentially missing `\n` and/or whitespace characters).

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,63 @@
...
+    **kwargs,
+) -> int:
+    # I believe we should be able pass argument to the subprocess function
     # from the caller. So we have to inject stdout/stderr/universal_newlines
-    kwargs["stdout"] = subprocess.PIPE
-    kwargs["stderr"] = subprocess.STDOUT
+    if process_stdout_line is None:
+        kwargs["stdout"] = subprocess.DEVNULL
+    else:
+        kwargs["stdout"] = subprocess.PIPE
```
vizovitin, line 43: Shorter and more straightforward: `kwargs["stdout"] = subprocess.PIPE if process_stdout_line else subprocess.DEVNULL`

vizovitin (changes requested):

pleskdistup/common/src/util.py
ukablan-wpc: Prepended cmd[0] to the logging lines.

pleskdistup/common/src/dpkg.py
ukablan-wpc: I introduced `dpkg.DPKG_TEMPFAIL_RETRY` with default values - one can set this variable to influence retry behavior of all those funcs. But they can also send arguments with custom retry intervals. Depends on usability, if users of this code won't need it, then we can only depend on global variable to set the retry behavior.

pleskdistup/common/src/util.py
ukablan-wpc: Removed these lines

pleskdistup/common/src/dpkg.py
ukablan-wpc: I accumulate it only when explicitly wanted: `collect_stdout = True`

pleskdistup/common/src/util.py
ukablan-wpc: It seems minor as it was from the previous impl and noone complained about it (yet?)

pleskdistup/common/src/util.py
ukablan-wpc: `f"{cmd[0]}: {line}"`
Not sure about process pid, worth to add it?

pleskdistup/common/src/dpkg.py
```diff
@@ -5,8 +5,12 @@
 import re
 import subprocess
 import typing
+import time
 
 from . import files, util
+from . import log
```
SandakovMM, line 11: NIT: NIT:  I usually put imports from common on one line. However, this is just a nitpick, and you can ignore it if you prefer.

pleskdistup/common/src/util.py
```diff
@@ -9,33 +9,57 @@
...
+        process.communicate()
+        return process.returncode
+
+    while process.poll() is None:
+        if process_stdout_line is not None:
+            if not process.stdout:
+                raise RuntimeError(f"Cannot get process stdout of command {cmd!r}")
+            line = process.stdout.readline()
+            if line:
+                process_stdout_line(line.rstrip('\n'))
```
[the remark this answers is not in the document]
ukablan-wpc: Possibly `*.log(line)` printed extra newline in the logs. If that's not important and not stripping last newline is important I can do it old way.

pleskdistup/common/src/util.py
ukablan-wpc: Not removing last newline - it's `"".join(stdout)` now

pleskdistup/common/src/dpkg.py
```diff
@@ -122,9 +125,59 @@ def safely_install_packages(
...
+            },
+        )
+        if exit_code == 0:
+            break
+        if i >= len(tmpfail_retry_intervals) or not cant_get_lock:
+            raise subprocess.CalledProcessError(returncode=exit_code, cmd=apt_get_cmd)
+        log.info(f"{apt_get_cmd[0]} failed because lock is already held, will retry in {tmpfail_retry_intervals[i]} seconds..")
+        time.sleep(tmpfail_retry_intervals[i])
+        i += 1
+    return "\n".join(stdout)
```
vizovitin, line 174: "\n" => ""

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,55 @@
...
+        log.err(f"{cmd[0]}: {line}")
 
-    # I beleive we should be able pass argument to the subprocess function
-    # from the caller. So we have to inject stdout/stderr/universal_newlines
-    kwargs["stdout"] = subprocess.PIPE
-    kwargs["stderr"] = subprocess.STDOUT
+    exit_code = exec_get_output_streamed(cmd, proc_stdout, proc_stderr, **kwargs)
+    if exit_code != 0:
+        log.err(f"Command {cmd!r} failed with return code {exit_code}")
+        raise subprocess.CalledProcessError(returncode=exit_code, cmd=cmd, output="\n".join(stdout))
```
vizovitin, line 24: "\n" => ""

pleskdistup/common/src/util.py
vizovitin: PID is transient, and there's no real concurrent execution, so there's not much use in it.

My point was more about inconsistency of having stdout and stderr logged with different levels. So alternative solution would be to log everything with INFO, but with prefix like `stdout: ` and `stderr: `.

Alternatively, you could leave this as is, but the current solution:
1. adds extra ERROR lines to the log, which are not necessarily errors;
2. adds extra ERROR words on the console, which might spook users.

pleskdistup/common/src/dpkg.py
vizovitin: OK, let it be as is, although I don't condone setting global constants from elsewhere.

pleskdistup/common/src/util.py
vizovitin: Right. Let's not document anything because otherwise we **need** to document arguments. (Actually we don't.)

vizovitin (changes requested):

pleskdistup/common/src/util.py
ukablan-wpc: Added. But we better impose similar conditions on other functions or have a consistent approach. Then, any programmer, by default, would add similar docstrings (at least).

pleskdistup/common/src/util.py
```diff
@@ -9,31 +9,59 @@
...
+        log.info(f"stderr: {line}")
 
-    # I beleive we should be able pass argument to the subprocess function
-    # from the caller. So we have to inject stdout/stderr/universal_newlines
-    kwargs["stdout"] = subprocess.PIPE
-    kwargs["stderr"] = subprocess.STDOUT
+    exit_code = exec_get_output_streamed(cmd, proc_stdout, proc_stderr, **kwargs)
+    if exit_code != 0:
+        log.err(f"Command {cmd!r} failed with return code {exit_code}")
+        raise subprocess.CalledProcessError(returncode=exit_code, cmd=cmd)
```
ukablan-wpc, line 24: @vizovitin @SandakovMM I deliberately removed `output = ''.join(stdout)` in `CalledProcessError`. It's really a pain when stdout is very large and program exits with non-zero (i.e. the leapp binary); then we read same output twice. IMO it doesn't make sense (sending all stdout in exception) as well.

vizovitin: OK, although I don't really see a real problem here.

pleskdistup/common/src/dpkg.py
```diff
@@ -140,10 +140,10 @@ def process_stdout(line: str) -> None:
         if collect_stdout:
             nonlocal stdout
             stdout.append(line)
-        log.info(f"{apt_get_cmd[0]}: {line}")
+        log.info(f"stdout: {line}")
 
     def process_stderr(line: str) -> None:
-        log.err(f"{apt_get_cmd[0]}: {line}")
+        log.info(f"stderr: {line}")
```
vizovitin, line 146: Are you sure `\n` at the end of the `line` will not be logged as well? I believe `logging` module does it by default.

vizovitin (changes requested):

pleskdistup/common/src/dpkg.py
ukablan-wpc: It will, that's why I've `.rstrip`ped it before calling the callback. I thought you had no issue about this newline?

pleskdistup/common/src/util.py
ukablan-wpc: Main problem for me is reading the logs :)

pleskdistup/common/src/dpkg.py
```diff
@@ -140,10 +140,10 @@ def process_stdout(line: str) -> None:
         if collect_stdout:
             nonlocal stdout
             stdout.append(line)
-        log.info(f"stdout: {line}")
+        log.info("stdout: {}".format(line.rstrip('\n')))
```
vizovitin, line 143: Shouldn't this also have `to_stream=False`?

vizovitin (approved):
Please rebase before merge.

SandakovMM (approved):

PR #93, opened 2024-11-25
Add a preliminary checker to ensure the mount points in /etc/fstab are in the correct order

SandakovMM:


pleskdistup/common/src/mounts.py
```diff
@@ -0,0 +1,47 @@
...
+
+    mount_points_order = {}
+    with open(configpath, "r") as f:
+        for iter, line in enumerate(f.readlines()):
+            if line.startswith("#") or line == "\n" or line == "":
+                continue
+            mount_point = line.split()[1]
+            mount_points_order[mount_point] = iter
+
+    missorderings = []
```
ukablan-wpc, line 33: possssible typo. Also at line 695.

pleskdistup/common/src/mounts.py
```diff
@@ -0,0 +1,47 @@
...
+    for mount_point in mount_points_order.keys():
+        if mount_point == "/":
+            continue
+
+        if mount_points_order["/"] > mount_points_order[mount_point]:
+            missorderings.append(("/", mount_point))
+
+        leading_paths = mount_point.split("/")[:-1]
+        for i in range(1, len(leading_paths) + 1):
+            parent_dir = "/".join(leading_paths[:i])
```
ukablan-wpc, line 43: Why not utilize `os.path.dirname(..)` loop until the parent and itself are equal?

SandakovMM: great idea actually

pleskdistup/common/src/mounts.py
```diff
@@ -0,0 +1,47 @@
...
+                continue
+            mount_point = line.split()[1]
+            mount_points_order[mount_point] = iter
+
+    misorderings = []
+    for mount_point in mount_points_order.keys():
+        if mount_point == "/":
+            continue
+
+        if mount_points_order["/"] > mount_points_order[mount_point]:
```
ukablan-wpc, line 38: Is it impossible to have no `/` root entry?

pleskdistup/common/src/mounts.py
```diff
@@ -0,0 +1,47 @@
...
+            misorderings.append(("/", mount_point))
+
+        parant_dir = os.path.dirname(mount_point)
+        while parant_dir != "/":
+            if parant_dir in mount_points_order and mount_points_order[parant_dir] > mount_points_order[mount_point]:
+                misorderings.append((parant_dir, mount_point))
+            parant_dir = os.path.dirname(parant_dir)
```
ukablan-wpc, line 45: `parent_dir`

pleskdistup/common/src/mounts.py
SandakovMM: I expect that a `/` will usually be inside the file, however, it is not necessarily required as far as I know. So let's add a test and fix it.
Thanks for pointing it out.

pleskdistup/common/src/mounts.py
```diff
@@ -0,0 +1,48 @@
...
+        its corresponding mount point.
+    Example:
+        >>> get_fstab_configuration_misorderings('/etc/fstab')
+        [('/home', '/home/user'), ('/var', '/var/log')]
+    """
+
+    if not os.path.exists(configpath):
+        return []
+
+    mount_points_order: typing.Dict[str, int] = {}
```
ukablan-wpc, line 25: Just curious, is it still beneficial to give type hints for local variables?

pleskdistup/common/src/mounts.py
```diff
@@ -0,0 +1,48 @@
...
+    with open(configpath, "r") as f:
+        for iter, line in enumerate(f.readlines()):
+            if line.startswith("#") or line == "\n" or line == "":
+                continue
+            mount_point = line.split()[1]
+            mount_points_order[mount_point] = iter
+
+    misorderings: typing.List[typing.Tuple[str, str]] = []
```
ukablan-wpc, line 33: Another question, since this is a return value the type hint is duplicated; is that worth?

pleskdistup/common/src/mounts.py
SandakovMM: I think so, since because we automatically check types using [the mypy GitHub action](https://github.com/plesk/dist-upgrader/blob/main/.github/workflows/mypy.yml)

pleskdistup/common/src/mounts.py
SandakovMM: Yes, we can remove it in this instance. However, I would like to retain the type because we have it for other local variables.

ukablan-wpc (approved):
lgtm

PR #95, opened 2024-12-09
Check fstab raid names

SandakovMM:


pleskdistup/actions/filesystem.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2023-2024. WebPros International GmbH. All rights reserved.
```
ukablan-wpc, line 1: Why not define in `mounts.py`?

ukablan-wpc: And then move `get_fstab_configuration_misorderings()` inside this source?

SandakovMM: > Maybe actions/mounts.py

Let's rename the file, why now.

> And then move get_fstab_configuration_misorderings() inside this source?

It might be challenging to import it from actions into the test, so I prefer to keep it in the common directory

pleskdistup/actions/mounts.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2023-2024. WebPros International GmbH. All rights reserved.
+
+import os
+
+from pleskdistup.common import action, mounts
+
```
ukablan-wpc, line 6: I think we can move `get_fstab_configuration_misorderings()` to here. Is there an obstacle that I can't see?

ukablan-wpc (approved):

pleskdistup/actions/mounts.py
SandakovMM: We will have to redo tests imports, which I would like to avoid. So I propose to keep it as is

PR #96, opened 2024-12-10
Check if repository file exsists before extracting data from it

SandakovMM:


pleskdistup/common/src/rpm.py
```diff
@@ -45,6 +45,9 @@ def extract_repodata(
     mirrorlist: typing.Optional[str] = None
     additional: typing.List[str] = []
 
+    if not os.path.exists(repofile):
+        raise FileNotFoundError(f"The repository file {repofile!r} does not exist")
```
ukablan-wpc, line 49: Seems like, it should give the same result as before:

```
$ python3
Python 3.6.8 (default, Nov 14 2023, 16:29:52) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> with open('eeree.txt', "r") as repo: 
...   l = repo.readlines()
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: 'eeree.txt'
>>> 
```
is it for log readability?

SandakovMM: > is it for log readability?

yep

PR #97, opened 2024-12-20
Fix mariadb version parsing

SandakovMM:


pleskdistup/common/tests/mariadbtests.py
```diff
@@ -17,6 +17,12 @@ def test_parse_utility_output(self):
         self.assertEqual(php.minor, 6)
         self.assertEqual(php.patch, 12)
 
+    def test_parse_utility_output_no_distrib(self):
+        php = mariadb.MariaDBVersion("mariadb from 11.6.2-MariaDB, client 15.2 for Linux (x86_64) using  EditLine wrapper")
```
ukablan-wpc, line 21: naming `vers` or `mariadb`?

pleskdistup/common/src/mariadb.py
```diff
@@ -56,6 +56,13 @@ def _extract_from_mysql_util(self, util_output: str):
         self.minor = int(minor_part)
         self.patch = int(patch_part)
 
+    def _extract_from_mariadb_util(self, util_output: str):
+        # String example: "mariadb from 11.6.2-MariaDB, client 15.2 for Linux (x86_64) using  EditLine wrapper"
+        major_part, minor_part, patch_part = util_output.split("from ")[1].split("-MariaDB")[0].split(".")[:3]
```
ukablan-wpc, line 61: Would it be more robust if we try to find the version with a regexp or maybe find first dot then track back the numbers...?

pleskdistup/common/tests/mariadbtests.py
```diff
@@ -6,16 +6,22 @@
...
+        self.assertEqual(mariadb_ver.major, 10)
+        self.assertEqual(mariadb_ver.minor, 6)
+        self.assertEqual(mariadb_ver.patch, 12)
 
     def test_parse_utility_output(self):
-        php = mariadb.MariaDBVersion("mysql  Ver 15.1 Distrib 10.6.12-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper")
-        self.assertEqual(php.major, 10)
-        self.assertEqual(php.minor, 6)
-        self.assertEqual(php.patch, 12)
+        mariadb_ver = mariadb.MariaDBVersion("mysql  Ver 15.1 Distrib 10.6.12-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper")
```
ukablan-wpc, line 15: Shouldn't this fail according to regexp?

SandakovMM: In this case, we use the `_extract_from_mysql_util` method instead of the new one, so the test case is fine

ukablan-wpc (approved):

PR #124, opened 2025-07-17
Add actions.PostRetrieveLicenseKeys

sofiiakulish:


SandakovMM (approved):

vizovitin (changes requested):

pleskdistup/actions/plesk.py
```diff
@@ -511,3 +511,36 @@ def _do_check(self) -> bool:
...
             raise
+
+
+class PostRetrieveLicenseKeys(action.ActiveAction):
+    """Some additional licenses, such as TuxCare ELS, should be regenerated for the new OS.
+\tRetrieving all keys may pose a risk, as some keys could become corrupted or invalid after conversion, which is not intended.
+\tAll keys are retrieved because it is currently not possible to retrieve the key only for a specific product using the CLI.
+\tChanges in the CLI would tie the implementation to the Plesk version.
+\tBy using this action, you accept the associated risk.
+    """
```
vizovitin, line 522:
```suggestion
    """ Some additional licenses, such as TuxCare ELS, should be regenerated for the new OS.

        Retrieving all keys may pose a risk, as some keys could become corrupted or invalid after conversion, which is not intended.
        All keys are retrieved because it is currently not possible to retrieve the key only for a specific product using the CLI.
        Changes in the CLI would tie the implementation to the Plesk version.
        By using this action, you accept the associated risk.
    """
```

vizovitin (changes requested):

pleskdistup/actions/plesk.py
```diff
@@ -511,3 +511,37 @@ def _do_check(self) -> bool:
         except Exception as e:
             log.err(f"Checking Plesk version has failed with error: {e}")
             raise
+
+
+class PostRetrieveLicenseKeys(action.ActiveAction):
+    """Some additional licenses, such as TuxCare ELS, should be regenerated for the new OS.
```
vizovitin, line 517: Space after `"""`.

vizovitin (approved):

PR #145, opened 2026-04-10
AlmaLinux 8 to 9 lib changes

ukablan-wpc:


pleskdistup/actions/common.py
```diff
@@ -212,6 +212,7 @@ class Reboot(action.ActiveAction):
     prepare_next_phase: typing.Optional[Phase]
     post_reboot: typing.Optional[action.RebootType]
     post_next_phase: typing.Optional[Phase]
+    do_before_post_reboot: typing.Callable[[], None]
```
SandakovMM, line 215: Hmm, why do we need this? I think we could simply add a separate action to the actions pool instead.

pleskdistup/actions/systemd.py
```diff
@@ -77,7 +82,8 @@ def _prepare_action(self) -> action.ActionResult:
         return action.ActionResult()
 
     def _post_action(self) -> action.ActionResult:
-        systemd.remove_systemd_service(self.service_name)
+        if self._remove_service_in_post:
```
SandakovMM, line 85: What’s the reason for not removing the service? If multiple reboots are required during the conversion, we could simply move this action to occur after the final reboot in the actions pool.

pleskdistup/common/src/leapp_configs.py
```diff
@@ -452,7 +503,10 @@ def create_leapp_vendor_repository_adoption(
         repofile: str,
         leapp_vendors_dir: str = LEAPP_VENDORS_DIR_PATH,
         ignore: typing.Optional[typing.List[str]] = None,
-        keep_id: bool = False
+        keep_id: bool = False,
```
SandakovMM: I think we could embed `keep_id` logic directly into `do_adapt_repository` using a closure or simple lambda/functor logic. That way, `new_repo` would either preserve the old `repo.id` or provide a new one as needed.

pleskdistup/common/src/leapp_configs.py
```diff
@@ -218,15 +219,47 @@ def adopt_repositories(repofile: str, ignore: typing.Optional[typing.List[str]]
...
+    files.rewrite_json_file(mapjson_path, mapping_json)
+
+
 def add_repositories_mapping(repofiles: typing.List[str], ignore: typing.Optional[typing.List[str]] = None,
                              leapp_repos_file_path: str = LEAPP_REPOS_FILE_PATH,
                              mapfile_path: str = LEAPP_MAP_FILE_PATH,
-                             skip_disabled: bool = False) -> None:
+                             skip_disabled: bool = False,
+                             do_adapt_repository: typing.Callable[[rpm.Repository, bool], rpm.Repository] = _get_adapted_repository,
+                             mapjson_path: str = "", distro: str = "",
```
SandakovMM: Let’s keep one argument per line

pleskdistup/common/src/leapp_configs.py
```diff
@@ -270,13 +304,20 @@ def add_repositories_mapping(repofiles: typing.List[str], ignore: typing.Optiona
...
-                    after_dist_repository = _write_repository_adoption(dist_repository_description, leapp_repos_file, False)
+                    after_dist_repository = do_adapt_repository(dist_repository_description, False)
+                    leapp_repos_file.write(repr(after_dist_repository))
                     if after_dist_repository is not None:
                         map_file.write("{oldrepo},{newrepo},{newrepo},all,all,x86_64,rpm,ga,ga\n".format(oldrepo=repo.id, newrepo=after_dist_repository.id))
 
                 leapp_repos_file.write("\n")
 
                 map_file.write("{oldrepo},{newrepo},{newrepo},all,all,x86_64,rpm,ga,ga\n".format(oldrepo=repo.id, newrepo=after_repository.id))
+                if mapjson_path:
```
SandakovMM, line 357: AFAIK, the mapping JSON file is just another way to configure mapping in leapp/elevate.
I’d suggest moving this logic into a separate function so each approach (json and csv files) can be used independently.
Splitting the function in two will also make it easier to add unit tests.

pleskdistup/actions/systemd.py
ukablan-wpc: It's because removing the service when the service itself is running causes term-like signals to be sent to the already running process(es) preventing finalization. Adding a "last second" action to remove the service could also be "solution" but it would be a wrong one since this action should be the very last although adding the service could be any step/action.

pleskdistup/actions/common.py
ukablan-wpc: Removing the service while the service itself is running, causes term-like signals to be sent to the already running service. We need to remove service right before reboot

pleskdistup/common/src/leapp_configs.py
ukablan-wpc: I am not sure what you mean here or whether it worths the effort. Can you give a concrete sample?

pleskdistup/actions/systemd.py
SandakovMM: I believe the script should not be stopped in this case, since it is declared as oneshot and uses RemainAfterExit=yes.

Optionally, we could move the service‑creation logic entirely out of the actions flow and handle it internally, giving us full control over its lifecycle.

For actions, I would prefer to keep a single unified flow so that when an action is marked as finished, we can be confident it has genuinely completed.

pleskdistup/actions/common.py
SandakovMM: I believe this issue should be resolved by marking the service as `oneshot` and using `RemainAfterExit=yes`.
However, this behavior may depend on how systemd handles the unit in practice. If the problem persists, I would suggest moving the systemd‑service creation logic out of the actions flow entirely and handling it internally within the dist‑upgrader.

pleskdistup/common/src/leapp_configs.py
SandakovMM: Something like:
```
from functools import partial

def get_adapted_repositorync(repository, keep_id):
    #...

f = partial(get_adapted_repositorync, keep_id=True)
```
However, let’s treat this as a NIT, since we previously had the keep_id anyway.

pleskdistup/common/src/leapp_configs.py
ukablan-wpc: AFAIR I saw a log about CSV mapping file is not regarded, only json (probably it's the new behavior). I will introduce another name function

pleskdistup/actions/systemd.py
ukablan-wpc: It's already 
```
        self._service_content = """
[Unit]
Description=Resume Plesk dist-upgrade after reboot
After=network.target network-online.target

[Service]
Type=oneshot
# want to run it once per boot time
RemainAfterExit=yes
ExecStart={util_path} --state-dir "{state_dir}" --resume {arguments}
TimeoutStartSec=infinity
TimeoutStopSec=infinity

[Install]
WantedBy=multi-user.target
"""
```

> Optionally, we could move the service‑creation logic entirely out of the actions flow and handle it internally, giving us full control over its lifecycle.

Then you need to make custom logic to delete the service before reboot after the post stage, no? I am not sure if it fits the design - it won't not generic flow runner any more.

> For actions, I would prefer to keep a single unified flow so that when an action is marked as finished, we can be confident it has genuinely completed.

Currently the action graph writer decides if anything should be run before post reboot. It doesn't force to be custom service removal or a housekeeping.

pleskdistup/actions/common.py
ukablan-wpc: Lets discuss in one thread only

pleskdistup/actions/systemd.py
SandakovMM: > Then you need to make custom logic to delete the service before reboot after the post stage, no?

Not only delete, but also create. So all the systemd‑service‑related logic is moved out of the actions‑flow scope.

pleskdistup/common/src/leapp_configs.py
SandakovMM: > AFAIR I saw a log about CSV mapping file is not regarded, only json (probably it's the new behavior). I will introduce another name function

It looks like they’ve fully switched to JSON mapping, although CSV still works for centos2alma. In any case, it’s good to have a dedicated JSON‑based function so we can migrate other tools to the new mapping approach as well.

pleskdistup/common/tests/leapp_configs_tests.py
```diff
@@ -1721,69 +1721,6 @@ def test_ignore_repository(self):
         self._compare_file_but_skip_empty(os.path.join(self.TEST_DIRECTORY, "ignore.repo"), expected_leapp_repos)
         self._compare_mapping_json(os.path.join(self.TEST_DIRECTORY, "ignore_map.json"), expected_mapping_json)
```
SandakovMM, line 1723: Ah, we were using keep_id before. My bad.
I’d suggest adding a test for `_get_adapted_repository` to ensure keep_id still behaves correctly on its side.

pleskdistup/common/src/leapp_configs.py
ukablan-wpc: Done

pleskdistup/common/src/leapp_configs.py
ukablan-wpc: Done

pleskdistup/common/src/leapp_configs.py
ukablan-wpc: Done

pleskdistup/actions/systemd.py
ukablan-wpc: Ok will check

pleskdistup/actions/packages.py
```diff
@@ -7,6 +7,19 @@
 from pleskdistup.common import action, dpkg, packages
 
 
+class AssertPackageIsNotInstalled(action.CheckAction):
```
SandakovMM, line 10: Hmm, I can't find this action used here or on the almalinux8to9 side. Do we need this action right now?

ukablan-wpc: This is hidden/collapsed due to change size I think:

https://github.com/plesk/almalinux8to9/pull/1/changes#diff-c0a7777243d3c868cae2b54f8ff0ee0951c3584ba89cfe5414bb38d7250bebdbR290

pleskdistup/actions/common_checks.py
```diff
@@ -1,14 +1,126 @@
...
+
+        duplicates = [repository for repository, count in collections.Counter(repositories).items() if count > 1]
+        if len(duplicates) == 0:
+            return True
+
+        self.description = self.description.format("\n\t- ".join(duplicates))
+        return False
+
+
+class AssertNoRepositoryDuplicates(action.CheckAction):
```
SandakovMM, line 16: Hmm, doesn't this action do the same thing as AssertThereIsNoRepositoryDuplicates? Should we keep only one of them?

pleskdistup/actions/common_checks.py
```diff
@@ -1,14 +1,126 @@
...
+    repofiles: typing.List[str]
+
+    def __init__(
+        self,
+        repofiles: typing.List[str] = files.find_files_case_insensitive("/etc/yum.repos.d", "*.repo"),
```
SandakovMM: I believe find_files_case_insensitive can potentially raise exceptions, and I'm not sure we handle that well enough during the actions initialization stage.
Could we move the file search logic into `_do_check`? We could just configure the target directory and regex here.

pleskdistup/actions/common_checks.py
ukablan-wpc: Done

pleskdistup/actions/common_checks.py
ukablan-wpc: But getting list of files from init is very natural and straightforward as well as keeps `_do_check` compliant with the interface. Hence, if it throws an exception (let's say not due to wrong type but directory traversal) it's at the very beginning of the program when nothing has started, no?
This is the cleanest approach I could think of, any other ideas?

pleskdistup/actions/systemd.py
ukablan-wpc: [PAUX-7064](https://webpros.atlassian.net/browse/PAUX-7064) [dist-upgrader] Move systemd continuation oneshot service management logic to core task runner

pleskdistup/actions/common_checks.py
SandakovMM: Hmm, maybe it's fine. Let's keep the call in __init__.
Also, AFAIK, in the current implementation we'll call `files.find_files_case_insensitive("/etc/yum.repos.d", "*.repo")` at the module import stage, not when `__init__` is called, which seems strange to me. I'd suggest using `None` as the default argument and comparing against it inside `__init__`.

SandakovMM (approved):

pleskdistup/actions/common_checks.py
ukablan-wpc: Done

PR #146, opened 2026-06-05
BUGFIX Leapp repo adaptation definition format 1.3.0

ukablan-wpc:


SandakovMM (approved):

pleskdistup/common/src/leapp_configs.py
```diff
@@ -728,6 +728,10 @@ def extract_leapp_report_inhibitors(json_report_path: str = "/var/log/leapp/leap
         inhibitors = _extract_leapp_report_inhibitors_from_txt(txt_report_path)
 
     if not inhibitors:
-        log.err(f"Neither JSON report file '{json_report_path}' nor text report file '{txt_report_path}' could be accessed. Please check if the files exist and have the correct permissions.")
+        if os.path.isfile(txt_report_path):
```
SandakovMM, line 731: It would be great adopt to new format if any, since usually report files are pretty huge. And also should be put in a feedback archive anyway.

PR #147, opened 2026-06-09
Add RemovePackages action to enable revertible remove packs

ukablan-wpc:


pleskdistup/actions/packages.py
```diff
@@ -99,6 +99,57 @@ def estimate_revert_time(self) -> int:
...
+    ):
+        self.packages = packages
+        self.tmpsavepath = tmpsavepath
+        self.name = display_name or f"removing packages {packages}"
+
+    def _prepare_action(self) -> action.ActionResult:
+        if os.path.isfile(self.tmpsavepath):
+            return action.ActionResult()
+
+        already_installed = packages.filter_installed_packages(self.packages)
```
SandakovMM, line 124: Nit: we could add type annotation here, since we have it for the same variable on the `_revert_action` side

SandakovMM (approved):

PR #152, opened 2026-06-22
yum.conf exclude repos operations to enable LEapp installation

ukablan-wpc:


pleskdistup/common/src/rpm.py
```diff
@@ -502,3 +503,48 @@ def get_package_installed_version(package_name: str) -> typing.Optional[str]:
...
+            return []
+    except OSError:
+        pass
+    return None
+
+
+def yum_conf_replace_exclude_list(new_excludes: str):
+    lines: typing.List[str] = []
+    replaced = False
+    with open(YUM_CONF_PATH, "r") as f:
```
SandakovMM, line 530: Should we check `if not os.path.exists(YUM_CONF_PATH):` the same way we do in the `yum_conf_get_exclude_list`?

SandakovMM (approved):

pleskdistup/common/src/rpm.py
ukablan-wpc: let's keep it as is for now - throw exception

PR #154, opened 2026-06-25
BUGFIX Opt to no hard fail on failing/non-existing GPG keys

ukablan-wpc:


pleskdistup/actions/rpm.py
```diff
@@ -53,11 +59,13 @@ def _prepare_action(self) -> action.ActionResult:
...
             except Exception as e:
-                log.err(f"Error occurred while fetching GPG key from '{key_url}': {e}")
-                raise RuntimeError(
-                    f"Unable to fetch GPG key from '{key_url}': {e}. To continue with the conversion,"
-                    f"please manually install the key into {self.leapp_gpg_keys_store!r}."
-                ) from e
+                log.warn(f"Error occurred while fetching GPG key from '{key_url}': {e}")
+                if self.is_hard_fail:
+                    raise RuntimeError(
+                        f"Unable to fetch GPG key from '{key_url}': {e}."
```
SandakovMM, line 65: Hm, why we want to avoid fail here? Could you please share more details about the issue you're facing?

ukablan-wpc: The main issue I want to solve is: https://github.com/plesk/almalinux8to9/issues/37 
The plan is checking installed packages from the caller and pass a flag here - just in case repo is found with no meaningful installation

SandakovMM: I would propose redefining `_is_required` on the `FetchImunifyGPGKey` side instead. I suppose we could skip this action since we consider a missing key not an error, so we likely believe the GPG key is not required for Leapp in this specific case.

ukablan-wpc: OK but this time we don't get error logs, my aim was to keep logs just in case

PR #155, opened 2026-06-30
BUGFIX Drain out stdout/err logs after poll is done

ukablan-wpc:


ukablan-wpc:
Discovered while fixing https://github.com/plesk/almalinux8to9/pull/41

SandakovMM (approved):

PR #160, opened 2026-07-21
BUGFIX Activate proftpd.socket if exists

ukablan-wpc:


pleskdistup/actions/systemd.py
```diff
@@ -254,6 +254,14 @@ def _enable_services(self) -> action.ActionResult:
 
         util.logged_check_call(["/usr/bin/systemctl", "enable"] + self.plesk_basic_services)
         util.logged_check_call(["/usr/bin/systemctl", "start"] + self.plesk_basic_services)
+        proftpd_socket_exists = False
+        try:
+            util.log_outputs_check_call(["/usr/bin/systemctl", "cat", "proftpd.socket"])
```
SandakovMM: I would propose use `systemd.is_service_exists` from `dist-upgrader/pleskdistup/common/src/systemd.py` here.

pleskdistup/actions/systemd.py
```diff
@@ -254,6 +254,14 @@ def _enable_services(self) -> action.ActionResult:
...
+            proftpd_socket_exists = True
+            util.log_outputs_check_call(["/usr/bin/systemctl", "enable", "proftpd.socket"])
```
SandakovMM: And `systemd.enable_services` here

pleskdistup/actions/systemd.py
```diff
@@ -254,6 +254,14 @@ def _enable_services(self) -> action.ActionResult:
 
         util.logged_check_call(["/usr/bin/systemctl", "enable"] + self.plesk_basic_services)
```
SandakovMM: Maybe even move whole action to `systemd` module function, if you have a time

pleskdistup/actions/systemd.py
ukablan-wpc: Done, thanks

pleskdistup/actions/systemd.py
ukablan-wpc: Yep makes sense

SandakovMM (approved):

sunkiepukgie-oss (approved):

PR #165, opened 2026-07-24
Add an action to pre-check if no MariaDB repository is enabled

SandakovMM:
In this case conversion will fail at the leapp preupgrade stage because leapp will be unable to find an appropriate mapping for MariaDB packages.

### DIFF
--- a/pleskdistup/common/src/rpm.py
+++ b/pleskdistup/common/src/rpm.py
@@ -525,6 +525,20 @@ def is_repository_url_enabled(repository_url: str, sources_dir: str = "/etc/yum.
     return False
 
 
+def has_enabled_repository(repofile: str) -> bool:
+    """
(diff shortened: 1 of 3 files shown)


pleskdistup/common/src/rpm.py
```diff
@@ -525,6 +525,21 @@ def is_repository_url_enabled(repository_url: str, sources_dir: str = "/etc/yum.
     return False
 
 
+def has_enabled_repository(repofiles: typing.Iterable[str]) -> bool:
```
ukablan-wpc: I would like to have this utility function to query only one repo file rather than a list

SandakovMM: Hm, can't see much difference, but why not. let's make it receive one repo file

ukablan-wpc (approved):

PR #168, opened 2026-07-28
Move leapp installation action into the dist-upgrader framework

SandakovMM:
The action should cover installation for all centos2alma, cloudlinux7to8 and almalinux8to9. Install required leapp packages passed as aruments and remove them on finishing stage and revert

### DIFF
--- a/pleskdistup/actions/leapp.py
+++ b/pleskdistup/actions/leapp.py
@@ -1,0 +1,162 @@
+# Copyright 2023-2026. WebPros International GmbH. All rights reserved.
+
+from pleskdistup.common import action, files, rpm, util
+
+import glob
+import os
+import shutil
+import subprocess
+import typing
+
+
+class LeappInstallation(action.ActiveAction):
+    """Install a fixed set of leapp packages from the ELevate repository.
+
+    Every RHEL-family conversion needs the same thing: add the elevate-release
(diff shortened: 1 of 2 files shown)


pleskdistup/actions/leapp.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def _get_installed_leapp_packages(self) -> typing.List[str]:
+        try:
+            return [name for name, _ in rpm.get_installed_packages_list(self.leapp_packages_glob)]
+        except subprocess.CalledProcessError:
+            # 'rpm -qa <glob>' behaviour on a glob matching nothing is not consistent
+            # between rpm versions, so treat a query failure as 'nothing is installed'.
+            return []
+
+    def _remove_leapp_packages(self, also_remove: typing.Optional[typing.List[str]] = None) -> None:
```
ukablan-wpc, line 71: why not default also_remove = [] ?

SandakovMM: AFAIK in Python we avoid passing array objects as a default value, because default objects are global. So if a function changes the object, it will be in a changed state on the next call. Right now we do not change it in `_remove_leapp_packages`, however I would like to avoid the antipattern anyway.

ukablan-wpc (approved):

PR #169, opened 2026-07-30
Add utility get_packages_with_sign_method

ukablan-wpc:


pleskdistup/common/src/systemd.py
```diff
@@ -84,6 +84,10 @@ def is_service_startable(
     return True
 
 
+def reexec_systemd_daemon():
```
SandakovMM, line 87: NIT: I would propose to split this change into a separate commit, because we have only commit "Add utility get_packages_with_sign_method - #169"

SandakovMM (approved):

pleskdistup/common/src/systemd.py
ukablan-wpc: Done

